Issue #39, opened 2020-07-13
Responsive left menu


buckyroberts:
Early analytics are showing that the majority of the users are coming in from mobile.

<img width="346" alt="Screen Shot 2020-07-13 at 4 40 04 PM" src="https://user-images.githubusercontent.com/8547538/87364111-973d6100-c527-11ea-9063-b105f003c047.png">

<img width="585" alt="Screen Shot 2020-07-13 at 4 37 54 PM" src="https://user-images.githubusercontent.com/8547538/87364004-51809880-c527-11ea-9247-0deb29c48734.png">

jacobclevenger:
The menu currently does not respond to the browser being resized. Consider adding a denounced window resize watcher to close the menu if the screen is small enough or open it if the screen is large enough.

Issue #30, opened 2020-07-13
Responsive top navigation/header


buckyroberts:
Responsive top navigation/header needed

<img width="759" alt="Screen Shot 2020-07-13 at 3 37 56 PM" src="https://user-images.githubusercontent.com/8547538/87360544-34e06280-c51f-11ea-8105-f6b518bdde49.png">

PR #41, opened 2020-07-14
Added responsive left menu navigation.

jacobclevenger:
Annotation 2020-07-13 233239

Annotation 2020-07-13 233313

### DIFF
--- a/src/containers/AdminLayout/AdminLayout.scss
+++ b/src/containers/AdminLayout/AdminLayout.scss
@@ -16,11 +16,40 @@
     grid-column: 1 / span 1;
     grid-row: 2 / span 1;
     overflow-y: auto;
+    
+    @media (max-width: 768px) {
+      &.isOpen {
+        transform: translateX(0);
+      }
+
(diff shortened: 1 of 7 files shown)


buckyroberts:
@jacobclevenger there is an interested situation that I didn't think of before that I just noticed when reviewing your branch. All of the pages need to support a responsive top menu (main site nav) however some pages also need to support a responsive left menu. I am not sure where to put the toggles for both of those or how they would look/work together. 

We have these issues in the design project for this repo:
- https://github.com/thenewboston-developers/Website/issues/30 - top nav/header
- https://github.com/thenewboston-developers/Website/issues/39 - left menu

@kristykjlee we should probably work on these together to make sure the design can support both

Screen Shot 2020-07-14 at 9 36 37 AM
Screen Shot 2020-07-14 at 9 36 45 AM

jacobclevenger:
For the navigation links in the top nav it is pretty common to put a button on the right of the nav menu and have a dropdown come from that.

PR #88, opened 2020-07-20
fix scroll on route change

whitehox:
Fix for scroll issues on route change


angle943:
https://github.com/thenewboston-developers/Website/pull/91

Issue #70, opened 2020-07-18
Left-Sidebar


guptstagram:
If sidebar navigation overlaps the content instead of sliding it then it will look better I think because in mobile it totally slides the content. Or one option can be it should take up the whole width in the mobile device.

We can achieve the overlap thing by **position: absolute** and **some height**. And on the click of hamburger changing display or width property.

![image](https://user-images.githubusercontent.com/31503495/87844495-5bb1e680-c8db-11ea-8710-7f9bb52650f7.png)

To make the navbar **width:100%** and on the click of hamburger changing display or width property.

![image](https://user-images.githubusercontent.com/31503495/87844559-b21f2500-c8db-11ea-99d0-b312e42d453e.png)

In both the cases, I have given some **top margin** to the sidebar as **thenewboston** logo should be **visible**.

buckyroberts:
Closed in place of https://github.com/thenewboston-developers/Website/issues/94

PR #90, opened 2020-07-21
Pull/87 - Part of Left Menu Issue #70 resolved

angle943:
Part of Issue #70

### DIFF
--- a/src/containers/AdminLayout/AdminLayout.scss
+++ b/src/containers/AdminLayout/AdminLayout.scss
@@ -1,3 +1,5 @@
+$TopNav-height: 46px;
+
 .AdminLayout {
   display: grid;
   grid-template-columns: 0 auto;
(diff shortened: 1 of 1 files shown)


angle943:
@buckyroberts question about the future of this issue:

1) If we do want to change the LeftMenu behavior such that it overlaps the main content, do we want it to default to it being opened? This might a little weird when a user gets to the page, and some contents in the screen is hidden underneath the open `LeftMenu`.

I do think that the shifting of content is very jarring though

Niekon01:
I would like to use the example of my University website. When we are om Desktop, the left menu is open by default (which can also be closed) and does not overlap the content, but pushes it to the left. On small screen devices, the menu hover over the main content, and everytime it is opened, it takes up the entire space.

Screenshot_20200721-232537_Chrome

The menu closes when we click outside the menu items.

buckyroberts:
Closed until we can get final requirements/designs https://github.com/thenewboston-developers/Website/issues/94

PR #96, opened 2020-07-22
close mobile nav menu on page switch

PatNeedham:
On mobile, when switching to a different page via the left nav / hamburger menu, the menu remains open but I think it's expected behavior for it to close (so visitors don't have to manually close it in order to read what is on that next page). Adding `toggleLeftMenuOpen(false)` to the AdminLayout's 2nd useEffect (taking the pathname argument) successfully does the trick, but doesn't work well on desktop since you'd expect it to remain open. So that's why it exists inside an if statement taking the screen width into consideration. Screen recording demonstrating the change: https://imgur.com/a/ZvcJ48t

### DIFF
--- a/src/containers/AdminLayout/index.tsx
+++ b/src/containers/AdminLayout/index.tsx
@@ -21,6 +21,9 @@ export const AdminLayout: FC<ComponentProps> = ({left, right}) => {
 
   useEffect(() => {
     rightDiv.current?.scrollTo(0, 0);
+    if (window.innerWidth < 768) {
+      toggleLeftMenuOpen(false);
+    }
   }, [pathname]);
 
   return (


src/containers/AdminLayout/index.tsx
```diff
@@ -21,6 +21,9 @@ export const AdminLayout: FC<ComponentProps> = ({left, right}) => {
 
   useEffect(() => {
     rightDiv.current?.scrollTo(0, 0);
+    if (window.innerWidth < 768) {
```
PatNeedham, line 24: 768px cutoff ensures the nav menu remains open for iPads, but on a Blackberry Playbook for example (600px) it would close:
image

buckyroberts:
Closing since we are updating the entire left menu design/logic

PR #99, opened 2020-07-23
Updated Design, Layout, Readme, and new Components.

jacobclevenger:
**Changes**
- Updated Readme
- Updated Design
- Updated Layout
- Updated Navigation
- Updated Various Styles through out application

**New Components**
- App Button - A general purpose button that is styled to match the design guidelines. See file for options
- App Toolbar - A general purpose toolbar. Currently only used in top nav. 
- App External Link -  This component should be used when linking to external content as it provides necessary attributes on the link element
- App Logo - Contains thenewboston logo.

**Images**

_Updated Designs_
image
image
image

_Mobile_
image
image
image


angle943:
I like the updated README 👍

Btw, i'm seeing some issues with the top menu bar. Are you seeing this as well?
Screen Shot 2020-07-23 at 6 53 34 PM

jacobclevenger:
What browser are you using? I didn't run into this issue on my end.

src/App/App.tsx
```diff
@@ -0,0 +1,72 @@
...
+ */
+const UsePage = ({ component: Component, layout: Layout, ...rest }) => {
+  return (
+    <Route {...rest} render={(props) =>
+      <Layout {...props}>
+        <Component {...props} />
+      </Layout>
+    }/>
+  );
+}
```
angle943, line 31: This is not too friendly with Typescript, I will try to come up with an alternative

src/components/index.ts
```diff
@@ -0,0 +1,7 @@
+export { AppButton } from './Button';
+export { AppIcon } from './Icon';
+export { AppLogo } from './Logo';
+export { AppToolbar } from './Toolbar';
+export { AppLayout } from './Layout';
+export { AppExternalLink } from './ExternalLink';
+export { LeftMenu } from './LeftMenu';
```
angle943, line 7: @buckyroberts I really like this. We should implement this for both `components` and `containers`. An added benefit is that we can exclude from this list the components that are used by another component, but shouldn't be used by itself. ie: `LeftSubmenu`

PR #112, opened 2020-07-27
Resync docs

buckyroberts:


src/containers/Guide/GuideResyncing/index.tsx
```diff
@@ -2,53 +2,152 @@ import React, {FC, ReactNode} from 'react';
...
+        and therefore a connection is no longer needed. Any confirmation validators that do have the requesting bank set
+        as their most trusted bank will follow the guidance of that bank and sync to the new primary validator.
       </p>
     </>
   );
 
-  const renderUpgradeRequest = (): ReactNode => (
+  const renderResyncProcess = (): ReactNode => (
+    <>
+      <h1 className="DocContainer__page-title GuideIntroduction__page-title">Resync Process</h1>
```
angle943, line 48: :( I was trying to avoid this type of classes, because whoever changes the styling in either of the two classes will almost certainly have no idea it is also effecting this one here. Do you think having more than one `h1` is something we should accodomate for in our docs?

buckyroberts: I'll move this it's own page.

PR #144, opened 2020-08-11
Update index.tsx

muhammetarslan:
Bucky not involving the transactions. So his transaction Id's will always be 1. In the last tx Amy and Carl's id will be 3 and their balance will be 13 and 7.


buckyroberts:
@muhammetarslan beautiful, thanks!

PR #157, opened 2020-09-04
bank asgi deployment

fonar:


fonar:
Hi @buckyroberts sorry, probably that's not the optimal way of making docs different for bank and validator, but if you'll give me an idea on how to do it, i'm ready to fix.

PR #162, opened 2020-09-09
Fixed the squashed step indicator for mobile view

Niekon01:
Previous
Changed


buckyroberts:
Added it was still squishing slightly above the break point (~800px) then I noticed we could just add `flex-shrink: 0;` to prevent that on all screen sizes.

PR #226, opened 2020-09-22
Leaderboard and Tasks Repository Filter URLs

thesanjeevsharma:
I have refactored the code to use routing instead of doing it by prop drilling. If you'd like me to change it back I can do that as well.

Also, I don't code in Typescript so please let me know if I did any blunders.

### DIFF
--- a/src/components/RepositoryFilter/index.tsx
+++ b/src/components/RepositoryFilter/index.tsx
@@ -4,7 +4,7 @@ import clsx from 'clsx';
 
 import {REPOSITORIES} from 'constants/github';
 import {GenericVoidFunction} from 'types/generic';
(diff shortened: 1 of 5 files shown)


buckyroberts:
@thesanjeevsharma Thanks! I am actually going to bed soon, but I will review tomorrow. Do you have an account number? If so please post it here since there is a 10,000 point reward for this issue :-)

thesanjeevsharma:
Sure, thanks! I just made an account. Here's my account number:

644cdaf382c1f711fb05fb4328409a62713ecc7a3e83fa07427cc3539a8a23ac

angle943 (approved):

Issue #129, opened 2020-08-07
Set a 720px Max Width for Guide & APIs Page Text


buckyroberts:
- For Guide & APIs pages, we will now have a max width of 720px specifically for the written information.
- **Leave diagrams/image sizes the same width they already are** (because some diagrams are best viewed past 720px width)

Guide
https://www.figma.com/file/ChyT88EmP2Rmzx35I9erCL/Marketing-Site?node-id=55%3A0

PR #243, opened 2020-09-27
Set a 720px max width for guide apis page text 129

audous:
Issue #129 work. I'm not sure if the feature branch name matters and I did not find an API header. I was assuming that API meant the items under guide. This is my first attempt to contribute here so please let me know if there is anything I am missing or can do to make Code Reviews run smoothly in the future. Thank you all TNB. -Phil


angle943:
hey @audous what is your Account #?

PR #257, opened 2020-09-28
resolve page moved horizontally

csmtcy:
The reason it gets scrolled horizontally is because of that table width.
I made it to become scrollable by auto in CSS, which I think is a better way.

Result


angle943:
@csmtcy Hey! The animated image above is the *updated* version, right? The original does not have any horizontal scrolling, correct?

buckyroberts:
Fixed in https://github.com/thenewboston-developers/Website/pull/266

Issue #220, opened 2020-09-21
[iPhone SE] Page moved horizontally


webdeveloperswj:
**Bug Description**
Page is horizontally scrollable where it should not be.

Url : https://thenewboston.com/guide/introduction
Video : https://recordit.co/lX2ssTout2

csmtcy:
#257 submitted a pull request for resolving the horizontal scroll issue

lapstjup:
#266 submitted a pull request for resolving the horizontal movement of page in Docs.

webdeveloperswj:
Issue solved

PR #266, opened 2020-10-02
Added a word wrap mixin to fix horizontal page movement in Docs

lapstjup:
This is the fix for the issue #220 

**Cause of the issue**
Since the table is not a fixed one, the padding applied to the td elements make the total table width go beyond for devices like iPhone SE. 

**Proposed Solution** 
We can break the words inside last element of td. I have done this for devices whose max-width is less than 400px using media query.
Since cross browser compatibility is preferable - I have referred to [Word Wrap Mixin](https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/) for the same.

newBoston_Website_220_fix_video (1)

### DIFF
--- a/src/components/Table/TableVertical/TableVertical.scss
+++ b/src/components/Table/TableVertical/TableVertical.scss
@@ -1,4 +1,4 @@
-@import '../../../styles/mixins';
+@import 'src/styles/mixins';
 
 .TableVertical {
   border: 1px solid #e6e6e6;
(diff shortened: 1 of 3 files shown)


angle943 (approved):

angle943:
New issue created related to this PR: https://github.com/thenewboston-developers/Website/issues/267

angle943:
@lapstjup what is your account #?

PR #277, opened 2020-10-04
276 - Add padding to last MenuGroup item on docs page

kcrnac:
I added 10px padding to last item from menu only when that item is not expanded. 
- 10px was added to stay consistent with padding which is already defined on top first element of navigation.

I tried making SCSS more prettier but could not do much because of :last-of-type selector which was used - as there is no way to select last element with specific CSS class (If there is any better way to do this, please let me know)

After this change, there is a little bit more space under last item of navigation.

Screenshot_15

### DIFF
--- a/src/components/MenuGroup/MenuGroup.scss
+++ b/src/components/MenuGroup/MenuGroup.scss
@@ -3,9 +3,7 @@
   font-size: 16px;
 
   &:last-of-type {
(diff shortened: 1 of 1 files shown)


angle943 (approved):

angle943:
@kcrnac what is your account #?

Issue #211, opened 2020-09-21
[Internet Explorer] Site not working 


webdeveloperswj:
Version 11
https://thenewboston.com/
![image](https://user-images.githubusercontent.com/59204822/93766544-32dc0380-fc34-11ea-8916-d50377ac57f5.png)

guptstagram:
Can I take it?

angle943:
@webdeveloperswj @guptstagram is this still an issue?

webdeveloperswj:
Yes @angle943

zinoadidi:
can i attempt it?

Hristijan95:
For now (and I think forever) let's just add a message that IE is not supported.
I highly doubt that someone will need to open this website from IE, ever.

nparoski:
I've created potential fix for this issue #1044

PR #281, opened 2020-10-05
Added IE 11 support using react-app-polyfill

itsnikhil:
### Fix for issue #211 [Internet Explorer] Site not working

- [x] Added npm package - [react-app-polyfill](https://www.npmjs.com/package/react-app-polyfill) which ensure language features are present to support Internet Explorer
- [x] CSS variable fallback
- [x] X-UA-Compatible meta tag
- [x] IE has limited SVG support. Tried my best to fix svg without destroying existing look and feel in modern browsers like chrome

### Final result screenshot
image

### Still broken on IE
- [ ] Docs page

### Special instructions
- Delete your entire node_modules folder
- Run `npm install` to re-install the node_modules with polyfill
- Run npm start to run again


angle943:
@itsnikhil hey is this ready for review? Also, could you resolve the merge conflicts and pull in the latest changes?

itsnikhil:
Hi, @angle943

Before I resolve the conflicts, I saw you removing css variable fallback from on of my previous PR. Fallback is necessary for IE. Most of the conflicts are also because of that only. I wanted your "Go" / "No-Go" on that.

angle943:
@itsnikhil good point about IE, but currently we use a lot of features that are not supported by IE site-wide, so let's not add the fallbacks.

If later we do realize that we need to add IE support (although I doubt the demand for IE support will ever get that big), we need to do a site-wide update. Let's add in the fallbacks then (or rather, remove css custom properties entirely)

Issue #305, opened 2020-10-16
[All mobile devices] Section should be aligned inline with other section above it


webdeveloperswj:
Url : https://thenewboston.com/help
![image](https://user-images.githubusercontent.com/59204822/96268238-6ff2a600-0fe6-11eb-86eb-542c50e26d09.png)

akshat-fsociety:
@webdeveloperswj can I work on this issue.

PR #307, opened 2020-10-18
Help page bug

GermaVinsmoke:
Solves issue https://github.com/thenewboston-developers/Website/issues/305
The problem was with the `align-items` property.


angle943 (approved):

buckyroberts:
@GermaVinsmoke Do you have an account number where we can send your coins to? Every accepted PR gets paid.

GermaVinsmoke:
I think my account number is this 8ff54e36f7fc534230faa1420c0116b2805abc4d8edb6a3661905ea0b4c875bd
What can we do with these points btw?

angle943:
@GermaVinsmoke think of it as cryptocurrency. Doesn't have too much value now, but it will increase as we increase in usage

GermaVinsmoke:
@angle943 This one is also not paid.

PR #306, opened 2020-10-18
Download page Linux

GermaVinsmoke:
Solves issue https://github.com/thenewboston-developers/Website/issues/301 
The code wrap problem

PR #309, opened 2020-10-20
Download Linux page

GermaVinsmoke:
There was some extra code which was not required at all in https://github.com/thenewboston-developers/Website/pull/306


angle943:
@GermaVinsmoke what's your TNB account #?

Issue #293, opened 2020-10-10
Add the functionality to use font-weight with our Inter Font


angle943:
Currently, we can't set the font-weight with our Inter fonts. We have to change the font-family to make it bold. For example: `font-family: var(--font-family-inter-bold)`.

Research into how to update this. We don't want to update the font weight by changing the font-family.

colfax23:
I'll take this one!

colfax23:
Ok, so I did some digging and it seems the approach outlined in this post is best:
https://www.456bereastreet.com/archive/201012/font-face_tip_define_font-weight_and_font-style_to_keep_your_css_simple/

This SO post shows a snippet of the implementation at a glance:
https://stackoverflow.com/a/10046346

I'll work on a PR the fonts file (https://github.com/thenewboston-developers/Website/blob/development/src/styles/_font.scss), leaving the existing font-family options in tact for to not break anything, and once that gets merged I can work on migrating the css to utilize the new font weight specification if you want!

angle943:
@colfax23 here's the AccountManager's same issue: https://github.com/thenewboston-developers/Account-Manager/issues/416

Open for you to take this one if you want ;). These two tickets alone will put you as #3 contributor on our leaderboard ;)

colfax23:
Consider it done!

PR #313, opened 2020-10-23
Utilize font-weight for Inter fonts

colfax23:
Hey all, this is the fix for #293 - I implemented the strategy I mentioned there.

The mapping I created for [inter style weight class](https://github.com/rsms/inter#design) to CSS font-weight is as follows:

Inter Style -- CSS font-weight

light -- 300
regular -- normal (400)
medium -- bold (700)
bold -- 800

I tested a few usages and it works.  If it's what you're expecting, we can merge (no functional changes yet), I can help convert the existing font-family for the Inter fonts to use the Inter family and then properly utilize the new font-weight, so we can deprecate the current usage.

Oh, and here's my account number :)
3cdb51ef0b7da3db65f47c170ec1815f3409c9781d4d8ca6c976273a732c9218

Thanks!

### DIFF
--- a/src/components/ContributorTasks/ContributorTasks.scss
+++ b/src/components/ContributorTasks/ContributorTasks.scss
@@ -68,11 +68,6 @@
     }
   }
 
-  &__task-title {
-    font-family: Inter;
-    font-weight: normal;
-  }
-
   @media (max-width: 1400px) {
     // Repository
     td:nth-child(2),
--- a/src/components/ContributorTasks/ContributorTasks.scss
+++ b/src/components/ContributorTasks/ContributorTasks.scss
@@ -69,7 +69,7 @@
   }
 
   &__task-title {
-    font-family: var(--font-family-inter);
+    font-family: Inter;
     font-weight: normal;
   }
 
--- a/src/components/DocWrapper/DocContainer/DocContainer.scss
+++ b/src/components/DocWrapper/DocContainer/DocContainer.scss
@@ -18,8 +18,8 @@
   }
 
   &__page-title {
-    font-weight: 300;
     font-size: 36px;
+    font-weight: 300;
     line-height: 1.2;
     margin-bottom: 24px;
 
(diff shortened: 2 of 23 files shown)


angle943:
this is great! can you update this PR to fulfill the following acceptance criterias:

1) Remove the old `Inter` stuff in the `/styles/_font.scss` file.
2) within `/styles/_font.scss`, the css Custom Properties (CSS varialbes) starting in line 70: all references of `--font-family-inter-...` should be removed. Inter will be our default font, so we should have no variables related to this.
3) Every instance of `--font-family-inter-bold`, `--font-family-inter-light`, `--font-family-inter-medium`, etc etc should be replaced with the `font-weight` counterpart.

colfax23:
Thanks!  Will do, I'll have it done in the next day or two.

colfax23:
Ok, updated!  I tested each update and there is no visual change, except for a very slight one for Task Title from the update to `src/containers/Tasks/TasksTask/TasksTask.scss` as previously the css was overriding a [font-weight of 600](https://github.com/thenewboston-developers/Website/blob/development/src/containers/Tasks/TasksTask/TasksTask.scss#L62) on top of the Inter Medium, which was landing the weight somewhere between Inter Medium and Inter Bold.  I could not reproduce the exact weight, but the difference is _very_ slight.

Let me know what you think.

angle943:
@colfax23 phenomenal work! I made a couple minor changes:

1) create a `--font-family-default: Inter, sans-serif`
2) for the body, have the style `font-family: var(--font-family-default)`
3) because of above, we no longer ever need to reference `Inter` in any of our other selectors, so I removed all other instances of `Inter`
4) make sure `font-size` comes before `font-weight` (alphabetize css)

If you plan on doing the same task for the AccountManager, please keep these in mind! But other than that, this is trully exactly what we wanted. Thanks!

colfax23:
Thanks for the feedback and making those changes!  Updated the AccountManager PR with those changes as well.

Issue #280, opened 2020-10-04
Update Account Manager Guide Pages


buckyroberts:
Update Account Manager guide screenshots that are out of date. For example:

## Old Account page

<img width="1409" alt="AccountOverview 9a07b5f3" src="https://user-images.githubusercontent.com/8547538/95025251-3791b600-063d-11eb-9c0b-5030c57b3171.png">

## New Account page

<img width="612" alt="Screen Shot 2020-10-04 at 12 27 44 PM" src="https://user-images.githubusercontent.com/8547538/95025242-29dc3080-063d-11eb-98a1-d93b9b5b7ad6.png">

colfax23:
I'll knock this out tomorrow!

PR #328, opened 2020-10-27
Updated guide images (#280)

colfax23:
Update for #280 

A few, observations/notes:
- For the Edit Nickname flow, there is an option shown in the screenshot "Add Signing Keys (for DevOps)" - I assume this is for pre release and will be removed for the prod version, just wanted to let you know it is in the new screenshot
- On that same flow, there seems to be a small bug, after you move your cursor off of that first option it does not get un-grayed out.  I was able to macgyver a screenshot with the hover appearing correct, but this might be worth a look.  I can create an issue if you want.
- Add Friends module, the spellcheck highlights the Account Number which also appears in the new screenshot I couldn't get it to go away


angle943:
1) ah yeah that's a good point. But I think it's fine for that DevOps stuff to be here. @buckyroberts please confirm.
2) Yeah that needs to be addressed. It's not a bug, its because we don't have a distinction between a focused state and a hover state (first button is focused, that's why it's that color. if you click 'Enter', it'll click the first button). We'll address that in a later design update. But thanks for pointing it out!
3) I just pushed a fix on the AccountManager, so there should no longer be a spellcheck highlight. Could you please retake that screenshot and update this branch?

I've also added a 1000 coin bounty on this Ticket

colfax23:
1. Cool, yeah just wanted to mention!
2. Ah, okay makes sense, np
3. Just updated

And awesome thanks!  I took this one despite not having a value yet because it seemed like it was important to get out before launch.  Thanks again, my address is: 3cdb51ef0b7da3db65f47c170ec1815f3409c9781d4d8ca6c976273a732c9218

Issue #351, opened 2020-11-01
Put "Hello World" in your OWN language!


angle943:
![helloworld](https://user-images.githubusercontent.com/5943963/97818023-d0275000-1c54-11eb-9c27-69099f35fd8d.png)

We currently have a text in the homepage where the "Hello World" loops through in different languages. We want you to add your own language!

in the file `/src/containers/Home/HomeHero/hello-world.ts`, you will find the `HelloWorld` object. The key must be in the two-letter ISO code of your language (ex: en. see https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes), and the value must be `Hello World` in your own language.

**Please open up a Pull request!**

You will get 500 coins for this. But limit one per person, and please only add a language if you are a native speaker!

And please write your account # on the PR, so that I can send you coins!

woosal1337:
Salam, Dünya!

angle943:
gotta open up a Pull request @woosal1337 !

woosal1337:
Which, branch?

Yamen2:
مرحبًا يا عالم 
Marhabaan ya-alealam

negoziator:
Hej verden 🇩🇰

DWK55:
hi im new to programming but have some experience i just dont know where src/containers/Home/HomeHero/hello-world.ts is to edit to "de"

Unlkoc:
> Which, branch?

Hello World in Turkish #480

Ramin-RX7:
Add Persian Hello World #477

Unlkoc:
> Hej verden 🇩🇰

Hello World in Turkish #500

gunars-b:
Added Hello World in Latvian

VJKLFT:
Added Hello World in Dutch #490

murattimurtas:
> ![helloworld](https://user-images.githubusercontent.com/5943963/97818023-d0275000-1c54-11eb-9c27-69099f35fd8d.png)
> 
> We currently have a text in the homepage where the "Hello World" loops through in different languages. We want you to add your own language!
> 
> in the file `/src/containers/Home/HomeHero/hello-world.ts`, you will find the `HelloWorld` object. The key must be in the two-letter ISO code of your language (ex: en. see https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes), and the value must be `Hello World` in your own language.
> 
> **Please open up a Pull request!**
> 
[...]

PR #361, opened 2020-11-02
Hello world in Hindi (India)

aayushi14:
#351


angle943:
paid!

Niekon01:
This is the translation of the phrase in Hindi. I think you can write "Hello World" as is in Hindi without translating since it is common.

Issue #371, opened 2020-11-02
[iPhone5/SE] Site moves horizontally


webdeveloperswj:
Url  : https://thenewboston.com/
http://recordit.co/yAVMuOr72D

Niekon01:
Does this problem still exist @webdeveloperswj

angle943:
this has been fixed i believe @Niekon01

Issue #369, opened 2020-11-02
Make the Footer's background color go all the way to the bottom


angle943:
On pages with less content, such as '/downloads', the footer's dark background color doesn't extend all the way to the bottom:

![Screen Shot 2020-11-01 at 9 56 20 PM](https://user-images.githubusercontent.com/5943963/97834821-97f03380-1c8d-11eb-9dae-4c3148cde439.png)

Make it such that this color stretches all the way to the bottom of the page, without arbitrarily giving the footer more height.

Like this:

![Screen Shot 2020-11-01 at 9 56 20 PM](https://user-images.githubusercontent.com/5943963/97835013-08975000-1c8e-11eb-9b2a-143c56f512b0.png)

lapstjup:
Hi @angle943 , When I go to the downloads page, there is no white space at bottom of footer. I am running it on macOS.

aayushi14:
@lapstjup To reproduce, check in iPad pro (1024x1366).

PR #375, opened 2020-11-02
Fix footer height and horizontal scroll for mobile devices

lapstjup:
This addresses two issues:-

1) Fix for #369 
2) Fix for #371


angle943:
#351

angle943:
@lapstjup hey sorry, but can we do one issues per PR? It makes payments too hard to deal with. You can just open a bunch of consecutive PRs next time. Also, #369 has been solved & closed

lapstjup:
@angle943 No worries. Also mention this in the slack group as well so that PR's ahead are not made in the same manner.

Issue #368, opened 2020-11-02
Upgrade A couple of our dependency libraries


angle943:
reference: https://github.com/thenewboston-developers/Account-Manager/pull/386/files

Upgrade the following packages (if they exist on the website)
![Screen Shot 2020-11-01 at 9 47 57 PM](https://user-images.githubusercontent.com/5943963/97834292-54e19080-1c8c-11eb-84ed-ab9e4c5b56e7.png)

When you finish, make sure the app doesn't break, and run `npm run lint`, and make sure there are no linting errors. If there are, fix them

amarjeet987:
👋

amarjeet987:
react-scripts require eslint version 6.6.0, so made that change. Rest of them are updated.

amarjeet987:
As per this, https://github.com/facebook/create-react-app/blob/master/CHANGELOG.md for eslint 7+ we need to update the react-scripts, let me know if I should proceed with this

PR #376, opened 2020-11-02
package updates

amarjeet987:
- react-scripts require a dependency of eslint 6.6.0, so updated it to that. Rest of them are updated as per the stated requirements.

#368

### DIFF
--- a/src/components/EmptyPage/index.tsx
+++ b/src/components/EmptyPage/index.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 
 import './EmptyPage.scss';
 
-const EmptyPage = () => {
+const EmptyPage = (): JSX.Element => {
   return (
     <div className="EmptyPage">
       <h1>No items to display</h1>
--- a/src/components/EmptyPage/index.tsx
+++ b/src/components/EmptyPage/index.tsx
@@ -1,8 +1,8 @@
-import React from 'react';
+import React, {FC} from 'react';
 
 import './EmptyPage.scss';
 
-const EmptyPage = (): JSX.Element => {
+const EmptyPage: FC = () => {
   return (
     <div className="EmptyPage">
       <h1>No items to display</h1>
--- a/src/components/LabelFilter/index.tsx
+++ b/src/components/LabelFilter/index.tsx
@@ -27,7 +27,7 @@ const LabelFilter: FC<ComponentProps> = ({className, handleLabelClick, selectedL
           className="LabelFilter__label"
           color={selectedLabelNames.includes(labelName) ? hexColor : 'e3e8ee'}
           key={labelName}
-          onClick={() => handleLabelClick(labelName)}
+          onClick={handleLabelClick(labelName)}
           name={labelName}
         />
       </div>
--- a/src/components/LabelFilter/index.tsx
+++ b/src/components/LabelFilter/index.tsx
@@ -6,7 +6,7 @@ import './LabelFilter.scss';
 
 interface ComponentProps {
   className?: string;
-  handleLabelClick: Function;
+  handleLabelClick: any;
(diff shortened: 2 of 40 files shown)


angle943 (changes requested):
@amarjeet987 yeah let's upgrade `react-scripts` too

tsconfig.json
```diff
@@ -14,7 +14,8 @@
     "resolveJsonModule": true,
     "skipLibCheck": true,
     "strict": true,
-    "target": "es5"
+    "target": "es5",
+    "noFallthroughCasesInSwitch": true
```
angle943, line 18: alphabetize keys ( 'n' should come before 't')

src/containers/PrimaryValidatorApi/PrimaryValidatorApiBankBlocks/index.tsx
```diff
@@ -40,7 +40,7 @@ const PrimaryValidatorApiBankBlocks: FC = () => {
 }`}
         heading="Request"
       />
-      <RequestResponseSnippet code={`{}`} heading="Response" />
+      <RequestResponseSnippet code="" heading="Response" />
```
angle943, line 43: we want these to be the way it was before. you might have to adjust eslint settings if it's complaining

angle943: Screen Shot 2020-11-03 at 8 18 21 AM

src/components/LabelFilter/index.tsx
```diff
@@ -6,7 +6,7 @@ import './LabelFilter.scss';
 
 interface ComponentProps {
   className?: string;
-  handleLabelClick: Function;
+  handleLabelClick: any;
```
angle943, line 9: handleLabelClick(labelName: string): () => void;

angle943 (changes requested):
if you run `npm run lint`, there are still issues that exist. Try to fix these without updating the eslint settings first. If it's not possible (nor reasonable), then you can add in the eslint rule changes. Feel free to message me about any rule you might not be certain about changing

Screen Shot 2020-11-03 at 8 20 14 AM

src/components/EmptyPage/index.tsx
```diff
@@ -1,8 +1,8 @@
-import React from 'react';
+import React, {FC} from 'react';
 
 import './EmptyPage.scss';
 
-const EmptyPage = (): JSX.Element => {
+const EmptyPage: FC = () => {
```
angle943, line 5: @amarjeet987 I actually like JSX.Element more, and in the future I do plan on removing the usage of `FC`, but for now, let's keep it as `FC` since that's what we use everywhere else

src/components/LabelFilter/index.tsx
```diff
@@ -27,7 +27,7 @@ const LabelFilter: FC<ComponentProps> = ({className, handleLabelClick, selectedL
           className="LabelFilter__label"
           color={selectedLabelNames.includes(labelName) ? hexColor : 'e3e8ee'}
           key={labelName}
-          onClick={() => handleLabelClick(labelName)}
+          onClick={handleLabelClick(labelName)}
```
angle943: the output of `handleLabelClick(labelName)` is itself a function

src/utils/toast.tsx
```diff
@@ -3,7 +3,7 @@ import {toast} from 'react-toastify';
 
 import Toast, {ToastType} from 'components/Toast';
 
-export const displayErrorToast = (error: string | {response: {data: string}; message: string}): void => {
+export const displayErrorToast = (error: any): void => {
```
angle943, line 6: let's keep this any for now

angle943 (approved):

Issue #355, opened 2020-11-01
iPad homepage scrolling issue


angle943:
when on the homepage on an iPad, there's two issues:

1) scrolling is not fluid
2) It appears that whenever the "Hello world" gets updated, the screen flicks back to the top.

Try to find a fix for these issues

angle943:
@lapstjup i will test this one once your changes are live. It was only happening on an actual iPad. If it's fixed, you'll be paid for this one as well

lapstjup:
@angle943 Were you able to test this?

angle943:
@lapstjup yup i just did! and it fixed it wonderfully. sending points your way

PR #443, opened 2020-11-04
Tap to scroll implementation

### LINKED ISSUE #402: iPhone clicking on top of the iphone doesn't scroll the website up
Usually, when you need to scroll up to the top of the page on an iphone, you click the top of your phone, and this should scroll the website up to the top. But on our website, it doesn't.

lapstjup:
Proposal to fix #402 

Changelog brief:- 

1) Removed the `overflow:hidden` from top level elements like `body` and `html` and removed height from Layout so that the sticky address bar minimises as user scrolls down and `window.scrollTo` api can work.
2) But removing the hidden property from them also means that the absolute positioned PopOvers and SlideUps have to be dealt in different manner. So to achieve that, I have made their parent div as fixed with appropriate `top` and `bottom` value so that the absolute positioning is done w.r.t to these fixed divs and not the `body` itself.

I have tried my best to test these changes so that nothing breaks.

Edit - This also might address #355


angle943:
@lapstjup this is truly fantastic work! You even took the portals into account. Very nice 👍

angle943:
note to self: possible fix for https://github.com/thenewboston-developers/Website/issues/355 ?

PR #458, opened 2020-11-05
replaced singlequotes with double quotes in lint command

### LINKED ISSUE #455: Make the command 'npm run lint' run for Windows/Linux users
currently, the command `npm run lint` doesn't work for Windows users (and possibly linux users). Lets update it such that it works for all platforms.

moki298:
should fix #455


angle943:
@moki298 let's use cross-env

PR #487, opened 2020-11-05
Bosnian Hello World

whitelionx:
fb84a32a698a03dd327c4ce673f19fa0acf6e68cb50f143d5d7c080471bc4625


shahraizali:
ISO Incorrect ❌ Source(https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)
Spellings Correct (it says hello world but google translate recognize it as Croatian maybe they are similar) ✅
~TO BE REVIEWED by @angle943

angle943:
@whitelionx can you put in the correct ISO code please? and also in alphabetical order

whitelionx:
I fixed the ISO code and yes they are identical as both come from the same family of languages (Balkan Slavic Languages) @shahraizali , also it should be alphabetically ordered correct I think @angle943

PR #494, opened 2020-11-05
Add `single-media-query` mixin for media queries

sno2:
The media queries were in-line and not using the `only screen` with them, therefore I created a mixin for simplifying this task.

idk if they are giving coins for small stuff but my account id is d84543947090d300910f8d1b9604c0c0a56fafca6cfbe52511c1e31757188bca


angle943:
@sno2 we are soon going to have fixed responsive viewports, and will have a couple mixins for those viewports. Closing this PR for now. We'll add those in once we create an `issue`. Keep on the lookout for that! I think we'll be writing up the issue fairly soon!

PR #499, opened 2020-11-06
Fixed Hello World in Korean

dwkim922:
The Korean was wrong, seemed like it was done with a translator. So I… changed it to the correct term, since Ive learnt programming in Korean before Im certain that this is right.

5020d46aaf94dd966a2e88b96ba7c143e3c778425b4329e49bac3d9c46829253


src/containers/Home/HomeHero/hello-world.ts
```diff
@@ -6,7 +6,7 @@ const HelloWorld: {[isoLangCode: string]: string} = {
   he: 'שלום עולם',
   hi: 'नमस्ते दुनिया',
   id: 'Halo Dunia',
-  ko: '여보세요 세계',
+  ko: '안녕 세계야',
```
angle943, line 9: @dwkim922 hey is it "안녕 세계야" or "안녕 세계" ?

dwkim922: Hey @angle943. 
Actually, "안녕 세상아" would sound better.

dwkim922: @angle943 should i pull request again with "안녕 세상아"?

angle943: it's cool, i'll update it on this PR once i get around it. thanks!

PR #540, opened 2020-11-06
Hello world in Slovenian

anzemur:
55fd2711036368e4081c1d5a4f76c8f08b0c8e5da2ca0e80c88f952b07ce317b


Niekon01:
ISO: wrong. The correct one should be "sl". (Source: https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)
Translation: correct
**Requires review**

Please put the ISO according to alphabetical order


@angle943

Issue #536, opened 2020-11-06
[Internet Explorer] Website Not Loading


webdeveloperswj:
**Bug Description**
Website not open In internet Explorer

**Steps to Reproduce**
1. Go to 'IE' (internet Explorer)
2.Enter Following Url "https://thenewboston.com/" 

**Expected behavior**
Website supports Every Browser 

**Actual behavior**
Website not supports internet explorer

**Screenshots/Recordings**
https://prnt.sc/vejidh

Source URL: https://thenewboston.com/
| Browser | Ie 11.0 |
| Operating System | Windows 10 |
| Resolution | 1920x880 |

zinoadidi:
I will fix this

thesanjeevsharma:
@webdeveloperswj @angle943 I think this issue is a duplicate of #211

zinoadidi:
#542 Pull request submitted

thesanjeevsharma:
Closing because this is a duplicate issue.

Issue #534, opened 2020-11-06
Following Devices And Browsers Website Not Loading 


webdeveloperswj:
**Bug Description**
Website Not Open Following Devices And Browsers

**Steps to Reproduce**
1.Following Devices And Browsers Go through

2) Websites not open 
Source URL: https://thenewboston.com/
| Device | iPad 4 |
| Browser | Tablet_safari |
| Operating System | iOS 7.0 |
| Resolution | 1536 x 2048 px |
| Screen size | 9.7 in - 5.8 x 7.8 in |
| Viewport | 768 x 1024 dp |
| Aspect Ratio | 4 : 3 |

3)Websites not Open
Source URL: https://thenewboston.com/
| Device | iPhone 6S Plus |
| Browser | Mobile_safari |
| Operating System | iOS 9.0 |
| Resolution | 1080 x 1920 px |
| Screen size | 5.5 in - 2.7 x 4.8 in |
| Viewport | 414 x 736 dp |
| Aspect Ratio | 16 : 9 |

screen shot :- https://prnt.sc/vegvos

4)Website not open
Source URL: https://thenewboston.com/
| Device | iPhone 6S |
| Browser | Mobile_safari |
| Operating System | iOS 9.0 |
| Resolution | 750 x 1334 px |
| Screen size | 4.7 in - 2.3 x 4.1 in |
| Viewport | 375 x 667 dp |
| Aspect Ratio | 16 : 9 |

screen shot :- https://prnt.sc/vegy1z

5)Website not open
Source URL: https://thenewboston.com/
| Device | iPhone 6 Plus |
| Browser | Mobile_safari |
| Operating System | iOS 8.0 |
| Resolution | 1080 x 1920 px |
| Screen size | 5.5 in - 2.7 x 4.8 in |
| Viewport | 414 x 736 dp |
| Aspect Ratio | 16 : 9 |

screen shot:-https://prnt.sc/vegzlb

6)Website not open
Source URL: https://thenewboston.com/
| Device | iPhone 6 |
| Browser | Mobile_safari |
| Operating System | iOS 8.0 |
| Resolution | 750 x 1334 px |
| Screen size | 4.7 in - 2.3 x 4.1 in |
| Viewport | 375 x 667 dp |
| Aspect Ratio | 16 : 9 |

screen shot:-https://prnt.sc/veh1o5

7)Website not open
Source URL: https://thenewboston.com/
| Device | iPhone 5S |
| Browser | Mobile_safari |
[...]

Issue #249, opened 2020-09-28
[iPad 4] Site not working


webdeveloperswj:
**OS iOS, v7
Browser  Safari**

![image](https://user-images.githubusercontent.com/59204822/94422163-46431d80-01a4-11eb-8c01-20a72cb3df3e.png)

angle943:
@webdeveloperswj does this issue still exist?

webdeveloperswj:
Yes @angle943

zinoadidi:
#542

PR #542, opened 2020-11-06
#211 - site now working Possible fix for related issues: (#536, #534, #249) 

zinoadidi:
The fix is tested on IE but the solution should apply across most browsers that are lacking behind on the latest JS API updates.

Account number 
e6877a1a3f9bfbe313c7d2022675c6cf409fcf4eb425793acec4f2d69cb86fcb

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "normalize.css": "^8.0.1",
     "qrcode": "^1.4.4",
     "react": "^16.13.1",
+    "react-app-polyfill": "^2.0.0",
     "react-cool-onclickoutside": "^1.5.5",
     "react-copy-to-clipboard": "^5.0.2",
     "react-dom": "^16.13.1",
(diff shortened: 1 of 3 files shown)


itsnikhil:
Hey,
I have already catered this issue #211 via https://github.com/thenewboston-developers/Website/pull/281 which was closed.
Sadly polyfills alone is not enough, css variable has been used all over the place along with SVG elements.
It was came to a conclusion to drop IE support.

zinoadidi:
cool. i guess we could at least show a message that it's not supported; right now its blank without a polyfill
With polyfill it works in some of the latest version at least;

itsnikhil:
Showing a message that IE is not supported would be better than a blank white page or a half baked broken page. 
Take approval from design team about this before you work on it

zinoadidi:
SO need to know who is on the design team that can give input about this so I can carry on with the task

itsnikhil:
@thesanjeevsharma @kristykjlee Can you please look into this

Hristijan95:
Let's just set a message that IE is not supported at the moment. I'll close this issue.

Issue #547, opened 2020-11-06
Teams page is not mobile responsive


thesanjeevsharma:
**Bug Description**
Teams page is not mobile responsive.

**Steps to Reproduce**
1. Go to [https://thenewboston.com/teams](https://thenewboston.com/teams) on any mobile device.

**Expected behavior**
Should be responsive on all screen sizes

**Actual behavior**
It is not responsive on mobile screens

**Screenshots/Recordings**

Checked in Browser
![image](https://user-images.githubusercontent.com/29539278/98371975-11f43400-2063-11eb-829a-053f116f8368.png)

Checked on Phone 
![Teams](https://user-images.githubusercontent.com/29539278/98372027-22a4aa00-2063-11eb-9c8b-45cd90f1d959.jpeg)

**OS and Browser**
 - OS: Android 10, Windows 10
 - Browser: Chrome, Firefox

**Account Number**
644cdaf382c1f711fb05fb4328409a62713ecc7a3e83fa07427cc3539a8a23ac

mersimoski:
Hi, i would like to take this, thanks.

mersimoski:
![image](https://user-images.githubusercontent.com/62712311/98413381-d1ec7b80-2079-11eb-93d8-7c20e10fea46.png)

GameOverTM0289:
I want to participate fixing this :D

jeremygblake:
I just used the Firefox mobile view option and it still is not responsive to the size of a mobile application.

BakungaBronson:
Hey guys. I'm willing to help out on this issue if an extra hand is needed.

Linzo99:
#566 : **Fixed responsiveness for teams page**

thesanjeevsharma:
I'll fix this. 😄 Change in design, we need vertical cards.

BakungaBronson:
I was thinking the same thing. On smaller devices, the text to the side just can't fit well without becoming too small, why not just put the text below the team member photo instead.

ArinzeJeffrey-droid:
I'd like to take this.

PR #553, opened 2020-11-06
Fixed responsiveness in TeamMemberCard

manishram:
This is a fix for #547
Reduced Avatar size in TeamMemberCard to make it responsive on phones


thesanjeevsharma (changes requested):
@manishram Your PR doesn't adhere to our guidelines. Don't club multiple issues in a single PR.

Also, your proposed solution for Team card responsiveness should include a css media query. With proposed solution, currently, width will decrease on desktop too.

Issue #461, opened 2020-11-05
User Account Prereq - Integrate our AccountManager's Modal into the website


angle943:
AccountManager's Modal:
https://github.com/thenewboston-developers/Account-Manager/blob/master/src/renderer/components/Modal/index.tsx

Figma reference:
https://www.figma.com/file/ChyT88EmP2Rmzx35I9erCL/Marketing-Site?node-id=1353%3A1915

![Screen Shot 2020-11-04 at 9 27 39 PM](https://user-images.githubusercontent.com/5943963/98201462-940a1e80-1ee4-11eb-8002-727e12c2fb76.png)

please incorporate this modal into the `Website`'s codebase, and install the necessary packages (such as Formik). Also you have to bring over a lot of the Form components, such as the Input. There will probably be util functions and hooks you gotta bring over

As a proof of concept, create the modal that is shown above. We don't have the page that contains the modal completed, so just make that modal pop up when you go to `/account`. The `Save` can just console log the inputs.

Majority of this will be a copy & paste job, but one thing different we are doing is giving the modal a responsive behavior.

![Screen Shot 2020-11-04 at 9 29 51 PM](https://user-images.githubusercontent.com/5943963/98201603-e2b7b880-1ee4-11eb-8404-7c435ad36d5a.png)

Update the `Modal` component such that it looks like the above for (max-width: 360px). (Please keep reusabability in mind. We want this Modal component to be as flexible as possible, so no hacky solutions.

amarjeet987:
I'll take it.

amarjeet987:
Made a PR: https://github.com/thenewboston-developers/Website/pull/585

Akshit1311:
Is it necessary to use Formik for this modal ?

itsnikhil:
![Web 1920 – 1](https://user-images.githubusercontent.com/20350674/99513641-dcc7da00-29b0-11eb-8bc7-60dc6aba83dc.png)

PR #585, opened 2020-11-07
Signup modal

amarjeet987:
fixes: #461 

- Added a test modal as per the UI given.
- Brought most of the components necessary from Account-Manager to the website for modals and forms.
- Installed necessary packages ( formik, react-select, @types/react-select ).
- Test modal can be accessed from route "/account".


Hristijan95:
Do we want all of the Formik components copied here, even though most of them are not being used right now? @thesanjeevsharma @lapstjup

charnysh-dev (commented):
@amarjeet987 
It looks good. but I'd like to know why we should have both `components/FormElements` and `components/FormComponents`. I think there is a way to use one of them or integrate them. What do you think? 
FYI @buckyroberts

amarjeet987:
@charnysh-dev  Yeah can be done ofc, I kept it that way to maintain the consistency between both the projects, account-manager and this. In account manager, we're already using stuff from both `components/FormElements` and `components/FormComponents`.

charnysh-dev (approved):

charnysh-dev:
> @charnysh-dev Yeah can be done ofc, I kept it that way to maintain the consistency between both the projects, account-manager and this. In account manager, we're already using stuff from both `components/FormElements` and `components/FormComponents`.

Ok, then we could resolve some conflicts

thesanjeevsharma (approved):

Issue #399, opened 2020-11-03
add the floating button that takes us to the top of the page


kristykjlee:

manishram:
@kristykjlee On which page we have to add that button on any particular or all pages?

manishram:
@kristykjlee I am working on it. Do you think the up facing thenewboston logo will be good?
![frefef](https://user-images.githubusercontent.com/22790904/98436242-d9566800-20ff-11eb-8104-7fe1d712873c.JPG)

mofodox:
@manishram I guess we can provide the icon yeah? 

cc: @kristykjlee

manishram:
Currently, I am using the upside logo of thenewboston as the "Go to Up" button. Let me know if you want change :)

![go-to-top](https://user-images.githubusercontent.com/22790904/98439827-7f16d080-211a-11eb-8ee7-caa2b5f58111.gif)

manishram:
@kristykjlee Please let me know about the design of **floating button**.

kristykjlee:
> @kristykjlee Please let me know about the design of **floating button**.

@manishram Heres the design for the floating button.

PR #600, opened 2020-11-07
Fixes: #399 Added go to top button

manishram:
Fixes: #399 

1. Floating button which takes the user to the top of the page when clicked.
2. Responsive both on phone and desktop
3. Floating button only visible after user scrolls down at least 100px

Working:

go-to-top-fix

Account: 0f92e975101eb96cea80dce430d74efd3172fa229377818d467ded85dcf2ab6c

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -19680,6 +19680,7 @@
           "integrity": "sha512-oWb1Z6mkHIskLzEJ/XWX0srkpkTQ7vaopMQkyaEIoq0fmtFVxOthb8cCxeT+p3ynTdkk/RZwbgG4brR5BeWECw==",
           "optional": true,
           "requires": {
+            "bindings": "^1.5.0",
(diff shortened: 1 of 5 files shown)


thesanjeevsharma (changes requested):
Fix these please:

1. css needs to be alphabetical order.
2. Don't import icon directly from mdi library. Use Icon component.
3. use useEffect for adding a listener and also do a cleanup for it. check react docs.
4. I don't understand why you rendered three icon components. 🤔 
5. Remove that unnecessary div wrapper.
6. Instead for GoToTop__go-to-top use GoToTop__button
7. Remove inline styles.

manishram:
Making chnages

manishram:
@thesanjeevsharma Made changes according to the requirements.

thesanjeevsharma (approved):

PR #624, opened 2020-11-08
fixes #619

### LINKED ISSUE #619: Docs need some language/grammatical fixes
**Bug Description**
There are not single but multiple instances of point renamed to coin where the meanings gets obscured. Also some grammatical nuances here and there.

**Places to fix as per my proof reading (Bold are incorrect and italics (if present) are correct) on the below routes**
1. **/guide/root-account-file :**
	 The root account file is a historic snapshot of all account balances at a given **coin** _point_ in time.
	 However, since Bucky had already verified the first transaction and also made a copy of the account file at that **coin** _point_ in time

2. **/guide/transaction-fees :**
	A very simple UI that Amy may see at that **coin** _point_ might look like this:

3. **/guide/banks :**
When users first download account manager applications and create their accounts, they will then select a bank **in** which they wish to use.

4. **/guide/confirmation-validators :**	•	
    1. This is the flattened representation of all account balances at a given **coin** _point_ in time
    2. Value representing the message hash of the primary **validators** _validator’s_ most recently verified block
    3. Beginning with the previous primary **validators** _validator’s_
	 
5. **/style-guide/css** - 
	1. if the naming **won't** won't be too confusing
	2. One of the pain-**coins** _points_ of BEM is the ambiguity
	3. This **coin** _point_ probably deserves a new section:
	4. Continuing on with the previous **coin** _point_

sdhiaeddine:
fixes #619

### DIFF
--- a/src/containers/Guide/GuideBanks/index.tsx
+++ b/src/containers/Guide/GuideBanks/index.tsx
@@ -35,7 +35,7 @@ const GuideBanks: FC = () => {
       <p>
         In addition to the responsibilities to the network, banks also have responsibilities to their users (account
         owners). When users first download account manager applications and create their accounts, they will then select
-        a bank in which they wish to use. They are then able to connect to the network through that bank and may begin
+        a bank which they wish to use. They are then able to connect to the network through that bank and may begin
         sending blocks. Users are able to send blocks through any bank of their choosing at any time. However, due to
         the fact that banks will typically only keep the transaction history of the accounts who have sent transactions
         through them, users will generally use the same bank for extended periods of time.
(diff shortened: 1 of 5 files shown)


Hristijan95:
Hi @sdhiaeddine , please share your account number so that we know where to send your coins. Every accepted PR gets paid. If you do not yet have an account then please download the account manager here https://thenewboston.com/download and follow these steps https://thenewboston.com/account-manager/create-an-account to create an account.

sdhiaeddine:
Hi @Hristijan95 , my account number: 119329b4d39d5483775efa0495487d04099bfce3fc6755345b5cc43da0b3611b
Thxs

PR #628, opened 2020-11-08
Update Mac download instructions to assist with open/security errors

zeuslawyer:
**What**
Update instructions for installing the Mac Account Manager App to assist users in resolving the security warning regarding installing the an application that "cannot be verified".

**Why**
Users experience confusion and concern with security notifications from the MacOS. Addressing them increases trust in TNB.

**Note**, this is my first PR so just taking something small :) Could not see an issue raised on this yet.

### DIFF
--- a/src/containers/Download/index.tsx
+++ b/src/containers/Download/index.tsx
@@ -91,7 +91,13 @@ const Download: FC = () => {
             <span className="instruction-container__instruction">Drag and drop the app to the Applications folder</span>
           </div>
           <div className="instruction-container__li">
-            <span className="instruction-container__instruction">Open the app</span>
+            <span className="instruction-container__instruction">
+              Open the app. If you see an error because the Account Manager app is not from the App Store check out{' '}
+              <a href="https://support.apple.com/en-us/HT202491" target="_blank" rel="noreferrer">
+                these instructions
+              </a>{' '}
+              to allow the install.
+            </span>
           </div>
         </>
       );


lapstjup:
@Hristijan95 Do we need this or are we working on a solution to prevent the security warning from popping up ?

zeuslawyer:
> @Hristijan95 Do we need this or are we working on a solution to prevent the security warning from popping up ?

the insertion can swiftly be removed once the update is done - consider whether the perception of unreliability does more damage to the TNB brand for people who arent technical but are exploring TNB.  Unless of course the patch for mac installer is coming out in the next day or two! :-)

Hristijan95:
@lapstjup Well, the current version of the acc manager still have the issue, so it wouldn't hurt to have instructions on the website (but we need to remember to remove them once the issue has been fixed).

There is an [issue in the Account Manager](https://github.com/thenewboston-developers/Account-Manager/issues/431) at the moment and people are working on it, but I am not sure how long will we need to get that fixed.

zeuslawyer:
agree @Hristijan95.  What is not a desirable outcome for visitors to TNB is for them to get blocked on install.  It may be worth considering leaving the instructions there even after the fix, as Mac's updates or a user's tight settings could cause that dialog box to pop up.  Ultimately anything that adds a friction point to a download should be avoided. 
cheers!

Hristijan95:
Let's add this, since there's a similar instruction for the windows installation. When we get rid of the problem, we'll remove the step.

Hristijan95:
Hi @zeuslawyer , is this PR connected to an issue? If yes, please link a reference to that issue here, as the rules require that every PR should be connected to an issue. If there is no issue, please create one and link it to this PR with a reference.

buckyroberts:
Fixes https://github.com/thenewboston-developers/Website/issues/1114

Rajputusman:
@zeuslawyer Reply with your TNB account number so we can send you your PR-reward coins. Thanks

PR #635, opened 2020-11-09
Added Hello World in Swahili account: 3972dfcfb9fd488cf90644b599558d3e265bdb883686933f62e8450aaee974b1

alexkimeu0:


Hristijan95:
@alexkimeu0 , there is already someone that translated Hello World in Swahili, but it's a different translation. So, which one is correct, "Habari Dunia", or "Salamu Dunia"?

alexkimeu0:
Oh. I didn't see it in the file. Both can be used.

Issue #648, opened 2020-11-09
teams.json ordering is not correct


lapstjup:
**Bug Description**
The ordering of team members in teams.json is not as per their contributor's id. Also there are some missing contributor ids. Need to order these entries as per the id's mentioned for consistency and prevent id overlapping issues during PRs. 

**Note**
 - If there's a miss (like 14), make sure that team member with id 15 changes to 14 , 16 changes to 15 and so on.
 - If there is a clash, chose any one to resolve it. 

We want to show the members in the same order on the [teams page](https://www.thenewboston.com/teams ) as we have been showing till now.

manishram:
@lapstjup I am working on it. Thanks

lapstjup:
Closing this since fix is done for the same directly in development branch.

PR #663, opened 2020-11-09
Fixes: #648 Reassigned id to duplicates

manishram:
Fixes: #648 
Assigned new id 37 and 38.


manishram:
@thesanjeevsharma should I assign id 14 and 16 instead of 37 and 38 because they have not been assigned to anyone yet.

manishram:
Bucky has already fixed the id. Closing this.

PR #690, opened 2020-11-10
Fixed a typo in Albanian language Hello World

Benett1:
I fixed a typo in the Albanian language, it was same as what google translate shows but there is a typo so this is the correct one.

account number: dfddf07ec15cbf363ecb52eedd7133b70b3ec896b488460bcecaba63e8e36be5


Hristijan95:

PR #694, opened 2020-11-10
Update hello-world.ts malayalam

KevinKurian7:
changed the malayalam text of hello world to its native phrase

### DIFF
--- a/src/containers/Home/HomeHero/hello-world.ts
+++ b/src/containers/Home/HomeHero/hello-world.ts
@@ -34,7 +34,7 @@ const HelloWorld: {[isoLangCode: string]: string} = {
   lt: 'Labas Pasauli',
   lv: 'Sveika pasaule',
   mk: 'Здраво Свету',
-  ml: 'ഹലോ വേൾഡ്',
+  ml: 'നമസ്കാരം ലോകമേ ',
   mr: 'नमस्कार जग',
   ne: 'नमस्कार संसार',
   nl: 'Hallo Wereld',


Hristijan95:
@KevinKurian7 , can you explain what do you mean but "native phrase"? Because we can't know which one is true, but according to google translate, the old one is the correct one, however I know that google translate can be wrong, so please explain what is what and I'll handle the issue.

aasifrafeeq:
in malayalam the meaning of  world is ലോകം

KevinKurian7:
> @KevinKurian7 , can you explain what do you mean but "native phrase"? Because we can't know which one is true, but according to google translate, the old one is the correct one, however I know that google translate can be wrong, so please explain what is what and I'll handle the issue.

what i meant by converted to native phrase  was that the text which was there for hello world was just literally saying hello world itself in Malayalam text but still as an English  word with the Malayalam alphabets
ഹലോ വേൾഡ്---original text ---pronounced hello world

നമസ്‌കാരം  ലോകമെ---text I suggested--
നമസ്‌കാരം  (pronounced  namaskaram meaning hello in English)
ലോകമെ       (pronounced lokame meaning world in malayalam)

It does not show up in google translate as you said when directly translating "hello world".

amarjeet987 (commented):
The initial one translates to "Hello World" and the new one to "Hello world", since we've the "World" in the website for the English text, idk let's go with that everywhere for the sake of consistency.

Hristijan95:
@KevinKurian7 , what is wrong with the existing one, what's incorrect?

PR #696, opened 2020-11-10
social-fix

rahulkiller104:
#686  fixed 
reduced font size for small device and remove left margin from social media to bring the card in the center in small devices
account number-dfddf07ec15cbf363ecb52eedd7133b70b3ec896b488460bcecaba63e8e36be5

Issue #686, opened 2020-11-10
"Join the Community" heading in /social


ddyakov:
**Bug Description**
It is not an actual bug, as it is more a styling problem.

**Steps to Reproduce**
1. Go to 'https://thenewboston.com/social'.
2. Toggle the screen to a mobile device resolution (e.g. from Chrome dev tools).
3. "Join the Community" is not aligned in the center.

**Expected behavior**
The heading should be centered the same way as on a tablet/desktop resolution.

**Actual behavior**
The heading is actually centered (because the parent div has 'justify-content: center;'), however, the font size is too much and it's making it look as if it is left aligned.

**Screenshots/Recordings**
![image](https://user-images.githubusercontent.com/11051204/98672776-9e9a3d00-2356-11eb-878f-e79315c364fd.png)

**OS and Browser**
 - Windows 10
 - Chrome

**Additional context**
The font-size might have to be less on mobile devices.

**Account Number**
f867dd69de4657143261b1f03f3e120e7cc4453ae8d77bce2a047d1989ef26ec

Hristijan95:
I also noticed that the Slack, Github, Youtube... cards are not centered correctly, they are leaning a little bit more to the right.

ddyakov:
@Hristijan95 Now that you said that, I double-checked and you are right.

T-Whiz:
Hey guys I managed to fix this by adding text-align: center; to the Social_heading class in the css.

From
.Social__heading {
    font-size: 36px;
    font-weight: 800;
    line-height: 1.25;
    margin: 0 auto 30px;
}

To
.Social__heading {
    font-size: 36px;
    font-weight: 800;
    line-height: 1.25;
    margin: 0 auto 30px;
    text-align: center;
}

![headerfix](https://user-images.githubusercontent.com/71679719/98730514-64d33180-236a-11eb-9a57-185b2824d9bf.png)

Hristijan95:
There are already 2 PRs opened about this issue, but both incomplete as of now.
@T-Whiz , that's great. If you fix the cards alignment as well, submit a PR and I'll close the issue.

T-Whiz:
I opened a pr but I'm not entirely sure I did it correctly, still a noob to github 👍 

I fixed the icons too by adjusting their margins. Both the fixes are super quick

PR #697, opened 2020-11-10
Fix uneven space for social buttons on mobile

kagisojamesthomba:
The social buttons were not evenly spaced in mobile view, there was more space on the left than on the right. I tested the following solution using chrome tools and reproduced the desired behavior. 1. Removed margin-left from "&__doc-button, .MarketingButton" class , 2. Changed justify-content in "&__button-container " class from center to space-evenly.

### DIFF
--- a/src/containers/Social/Social.scss
+++ b/src/containers/Social/Social.scss
@@ -30,6 +30,7 @@
     font-weight: 800;
     line-height: 1.25;
     margin: 0 auto 30px;
+    text-align: center;
   }
 
   &__subtext {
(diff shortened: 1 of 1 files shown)


Hristijan95:
Hi @kagisojamesthomba , thank you for your contribution. Is this PR connected to an issue? If yes, please link a reference to that issue here, as the rules require that every PR should be connected to an issue. If there is no issue, please create one and link it to this PR with a reference.

kagisojamesthomba:
bug #686

Hristijan95:
The text centering was not fixed, I fixed it, but there is still an issue on a screen ~458px width:

Screen Shot 2020-11-10 at 18 32 52

Also, this is a duplicate with #696 , so whoever commit a better solution, I will accept the PR.

Hristijan95:
Fixed in #716 - Thanks for the effort.

PR #698, opened 2020-11-10
Update of hello-world.ts and a suggestion.

phoenixArtwork:
.0.
Russian translation of 'Hello World' (with upper case 'W') is 'Привет Мир';
Russian Translation of 'Hello world' (with lower case 'w') is 'Привет мир';
Ukrainian translation of 'Hello World' (with upper case 'W') is 'Привіт Світе'
Ukrainian Translation of 'Hello world' (with lower case 'w') is 'Привіт світе'.
As the original line is 'Hello World', the translations should be:
uk: 'Привіт Світе',
ru: 'Привет Мир',
I believe the same upper/lower case mismatch is true for some other translations.

.1.
I suggest changing the original line 'Hello World' to 'Hello, World!' which is both a more grammatically correct and more energetic form.
If you choose to do so, here are the translations:
uk: 'Привіт, Світе!',
ru: 'Привет, Мир!',

My Account Number = 0e86adf963b7801823b68e2911ab7884eab022ef61ea8cd964eff00293a5418f


amarjeet987 (approved):
Lol nice, good observation. 👍

Hristijan95:

Issue #587, opened 2020-11-07
Dropdowns are not keyboard accessible


lapstjup:
**Bug Description**
The Get Started, Community and More dropdowns are not keyboard accessible.

**Steps to Reproduce**
Scenario 1:-
1. Go to [thenewboston](https://www.thenewboston.com/)
2. Click on Get Started or any other dropdown text.
3. Press Tab Key

Scenario2:-
1. Go to [thenewboston](https://www.thenewboston.com/)
2. Press tab key until you are over any of the dropdown text, press enter/space and press tab key again.

**Expected behavior**
The list of items of dropdown should get focus

**Actual behavior**
The list items do not get focused

**Screenshots/Recordings**
[Recording](https://drive.google.com/file/d/14BC82nRqS6t_F9aH7C6Ke_SS8JpjiYAP/view?usp=sharing)

**OS and Browser**
 - OS: [e.g. macOS]
 - Browser [e.g. Chrome, Firefox, Safari]

**Additional context**
Also multiple dropdowns can be opened at the same time using enter/space. This can be also be corrected in this same issue.

prajwalkulkarni:
I went through this issue and made some research and caught hold of a function that might be of use. I was unsure of the implementation, so here's the code so far. Please check if it works.
`
const useKeyPress = function (targetKey: any) {  //Not sure where to use this function.
    const [keyPressed, setKeyPressed] = useState(false);
    function downHandler(key: any) {
      if (key === targetKey) {
        setKeyPressed(true);
      }
    }
    const upHandler = (key: any) => {
      if (key === targetKey) {
        setKeyPressed(false);
      }
    };

    React.useEffect(() => {
      window.addEventListener('keydown', downHandler);
      window.addEventListener('keyup', upHandler);
      return () => {
        window.removeEventListener('keydown', downHandler);
        window.removeEventListener('keyup', upHandler);
      };
    });
    return keyPressed;
  };
 `

nparoski:
Since this is still unsolved I'll give it a go.

mbdteam9:
Hi @lapstjup, im sorry not sure about the complex code others have suggested. 

I just know how to use fake buttons(keyboard acessible) , by making use of tabindex=0. 

If it dont work sorry, im too new to web dev.

lapstjup:
Hey @mbdteam9 No need to be sorry for anything. You can always raise a PR for a fix you think can work and post review if we find it to be fit for inclusion, we will merge it 😄.

[...]

PR #726, opened 2020-11-11
issue#303 - Removed dotted outline on Mozilla Firefox and all browsers

mersimoski:
Added !important on the outline:none so that every browser will take it priority to not show outline.

acc id:
cf59a784ff69e64e977359f28cd4d04c0aea7a37cf1b8bda3dfbde2364032855


lapstjup:
Hey we need the dotted line for accessibility purposes. We already have an issue #587 which requires dropdown to be keyboard accessible. That also means ensuring that the dropdown button's are accessible as well and the outline helps in achieving that.

Also if you feel like there is a bug or a missing feature, create an issue for it first. We follow 1 issue - 1 PR.

PR #731, opened 2020-11-11
Adding hello world in Swedish

VebjornRiiser:
Account number: "cc67fd28b3daca5378b6af9650b719a1025f756ec0d17d4baf7e7e5487cab828"
issue: #351


Hristijan95:
The ISO code is incorrect, should be sv, not se ([see here](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)), and the Swedish language (sv) is already added, therefore I am closing this PR.

PR #732, opened 2020-11-11
added hello world in nepali language

volatileColors:
No account number cause I don't want points for this small commit. I only want to get points when I make some useful pull requests


Hristijan95:
The ISO code is incorrect, should be ne, not np ([see here](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes)), and the nepali language (ne) is already added, therefore I am closing this PR.

Issue #747, opened 2020-11-12
Sorting of team members not working as expected


manishram:
**Bug Description**
In `/src/data/teams.json` file all members got **"createdDate": "2020-07-08T02:14:59.307535Z"** and few got **"createdDate": "2020-05-09T02:14:59.307535Z"**. People who got the 2nd one had got their position misplaced and not sorted according to their **contributorId**.

**Steps to Reproduce**
Go to https://thenewboston.com/teams

**Expected behavior**
All members sorted on the basis of **contributorId**

**Actual behavior**
Few members got their position misplaced

**Screenshots/Recordings**
Member with  **contributorId** **5, 18 and 23** got their position misplaced.

![id](https://user-images.githubusercontent.com/22790904/98913102-f4a0e900-24ec-11eb-8b0f-6415b80f99f9.JPG)
![id-2](https://user-images.githubusercontent.com/22790904/98913125-f8cd0680-24ec-11eb-9558-0eb1d7af0e3c.JPG)

**Additional context**
This happened to those members who are in Back-End Developers. Few of them got the 2nd **createdDate**. The possible quick solution is to change their `createdDate` to `2020-07-08T02:14:59.307535Z` because all other members also got the same `createdDate` irrespective of when they joined the team. Or we need to correct the sorting.

**Account Number**
0f92e975101eb96cea80dce430d74efd3172fa229377818d467ded85dcf2ab6c

asaadat99:
Hello, I am interested in working on this, can you assign it to me?

Hristijan95:
@asaadat99 , as you can se above, there's already a PR to fix this from @manishram . However, at the moment the design team is doing a redesign on the Teams/All page, so we'll decide what to do with this when the redesign is ready.

jamessspanggg:
The issue is that the team members are sorted based on the `created_date`, instead of the `contributorId` as shown here:
https://github.com/thenewboston-developers/Website/blob/078f8d4cf949c544a9719d3fd88ebaf43c8d4174/src/utils/data.ts#L67-L73

Changing it should be fine. @manishram will you be working on it?

manishram:
Yeah sure @jamessspanggg go ahead. I am working on another issue right now. :)

PR #749, opened 2020-11-12
Quick Fix: Changed the createdDate of few members

manishram:
Fixes: #747
This is a quick fix by changing dates because `createdDate` have vague values.

"createdDate": "2020-05-09T02:14:59.307535Z" =>  "createdDate": "2020-07-08T02:14:59.307535Z"

`createdDate` for all members is the same irrespective of when created.


amarjeet987 (changes requested):
Thanks. Can you please also check why it's not getting sorted as per the contributorId and find a fix for it so that the issue is fixed once and for all. Do let me know if you need any help with that.

manishram:
#788 fixes the sorting.

Issue #564, opened 2020-11-07
Task Left Menu Issue


buckyroberts:
@kristykjlee we have a new repo that we want to include on the Tasks page on the website, however the name of the repo is `thenewboston-python-client` which is a bit long. 

<img width="856" alt="Screen Shot 2020-11-06 at 4 13 22 PM" src="https://user-images.githubusercontent.com/8547538/98425739-1f9dbf80-204b-11eb-9280-3087260d6d5b.png">

manishram:
Why not just replace thenewboston with TNB and write it like TNB-python-client?

![gregeg](https://user-images.githubusercontent.com/22790904/98432950-ffbcd900-20e8-11eb-995d-9939d4e1d02f.JPG)

mofodox:
And maybe the naming convention should be ‘TNB-Python’ and ‘TNB-Python-Client’?

myszon01:
what projects thenewboston-python-client and thenewboston-python are mainly responsible for?

stainlezzking:
since you already have a "thenewboston-python" that you're okay with, it's only symmetrical to name the other one "thenewboston-python-C" but if you're okay with changing both names, then I'll go with the idea of the guy above me  'TNB-python-client

Akash-S1999:
I would recommend changing "thenewboston" to TNB. Like "TNB-python", "TNB-python-Client" In that way, it will look more appealing and simple.  Also, we can skip the word python and include it in the documentation part. 
The latter idea is better as we can have a menu "TNB-python" and the subpart can be "TNB-python-client" or just "TNB-Client" as it will already be the subpart of "TNB-Python"

JanGoO-dev:
I think this will be a better solution, If you don't want to change your repository name.
Bellow Image is a basic demonstration of the solution. I mean you should give a common title to both of these two repositories and make it unclickable while bellow that you should have two small buttons for each of the repositories.

Example:
thenewboston
--python
--python-client

I hope you understand. English is not my native language that's why there might be some mistakes.

![Screenshot-2020-11-07-194401](https://user-images.githubusercontent.com/55313789/98444715-05401080-2135-11eb-8e02-d13e0c5210d1.png)

I am a big fan and student of your. I learned CPP by watching your videos and know I am studding computer science. You inspired me and your work always motivates me. Thank you for everything.

Ayush-Porwal:
@buckyroberts can I take this? I think this is easy and it would be a nice start for me. Thank you!

[...]

PR #803, opened 2020-11-13
Fixes Left Navigation Links

manishram:
Fixes: #564

Converted data inside `REPOSITORIES` `OPENING_CATEGORY_FILTERS` and `TEAM_NAME_FILTERS` into objects with `title` and `url` properties.

left

Account:
0f92e975101eb96cea80dce430d74efd3172fa229377818d467ded85dcf2ab6c


lapstjup:
@manishram It would be good if @Hristijan95 reviews your changes since I guess these might be requested by him. Although instead of `any`, it would be better to create an interface for the title,url structure. Also I think - title,pathname sounds better. Confirm the same with him.

Hristijan95:
Yep, I saw this yesterday and somehow forgot to comment. You need to create an interface instead of using any and bring back the generic <T> on the places where needed, I am not sure I understand why you changed all ComponentProps of FlatNavLinks to any. They need to be <T> & only create an new interface for the options type.
Also yeah, if it's no trouble, pathname sounds better.

src/utils/github.ts
```diff
@@ -4,8 +4,8 @@ import {AMOUNT_COLOR, REPOSITORIES} from 'constants/github';
 import {BaseRelease, Issue, Release} from 'types/github';
 
 export const fetchGithubIssues = async (): Promise<Issue[]> => {
-  const promises = REPOSITORIES.map((repoName) =>
-    axios.get(`https://api.github.com/repos/thenewboston-developers/${repoName}/issues`),
+  const promises = REPOSITORIES.map((repoName: any) =>
```
jamessspanggg, line 7: opps, can remove the `any` here

manishram: done

manishram:
This PR is in collaboration with @jamessspanggg . Thankyou James!!

lapstjup:
@manishram and @jamessspanggg I have done some refactoring. The reason behind it - FlatNavLink doesn't make sense to be generic anymore since it's internal implementation depends on pathname and url fields which is obtained from NavOption.

src/containers/Teams/index.tsx
```diff
@@ -1,32 +1,32 @@
...
+  {pathname: TeamName.frontEndDevelopers, title: 'Front-End Developers'},
+  {pathname: TeamName.kotlinSDK, title: 'Kotlin SDK'},
+  {pathname: TeamName.marketing, title: 'Marketing'},
+  {pathname: TeamName.payments, title: 'Payments'},
+  {pathname: TeamName.qa, title: 'QA'},
+  {pathname: TeamName.redditModerators, title: 'Reddit Moderators'},
+  {pathname: TeamName.research, title: 'Research'},
+  {pathname: TeamName.security, title: 'Security'},
+  {pathname: TeamName.slackManagers, title: 'Slack Manager'},
+  {pathname: TeamName.youtube, title: 'TouTube'},
```
jamessspanggg, line 29: Should be `YouTube`

jamessspanggg (changes requested):

Hristijan95:
Looks really nice, I'm merging it

PR #825, opened 2020-11-14
Added Hello World in Klingon

### LINKED ISSUE #351: Put "Hello World" in your OWN language!

at-in:
Resolves #351 

Klingon only has 3 letter ISO code and doesn't have a ISO 639-1 code(2 letter ISO code).

My account number: `814168a356e3f9046900838ee068fc87903e2fd47e8f749f84ca6780ec821059`

### DIFF
--- a/src/containers/Home/HomeHero/hello-world.ts
+++ b/src/containers/Home/HomeHero/hello-world.ts
@@ -76,6 +76,7 @@ const HelloWorld: {[isoLangCode: string]: string} = {
   te: 'నమస్తే ప్రపంచం',
   th: 'สวัสดีโลก',
   tl: 'Kamusta Mundo',
+  tlh:"qo' vIvan",
   tn: 'Dumelang Lefatshe',
   tr: 'Merhaba Dünya',
   uk: 'Привіт світе',


lapstjup:
This is a constructed language and not a native one. Even though this evolved from fictional to real world usage by many fans, it's still not the one approved for ISO-639-1 usage which was made to identify languages for computer systems.

at-in:
Thanks for the reply @lapstjup but will adding it cause any problem like any bug? If not than what is the problem in adding it since you said it has evolved to real world usage by many peoples.

PR #832, opened 2020-11-15
Add last modified label to leaderboard

### LINKED ISSUE #786: [Feature Request] Add last updated date time to leaderboard
Since the leaderboard is based on commits to the `tasks.json`, the leaderboard may not contain the latest updated results. It would be good if the last updated date time is shown in the leaderboard.

ffpetrovic:
Fixes #786


buckyroberts:
We are moving all the tasks/leaderboard data over to the API. Would prefer not to merge this in since we would just have to remove it soon.

PR #842, opened 2020-11-15
added hello world in averic

MrRamoun:
added hello world in averic to the hello-world.ts file


thesanjeevsharma:
@MrRamoun Thank you for your effort but Avaric has already been added!

PR #850, opened 2020-11-15
Fix download page to only fetch latest release

### LINKED ISSUE #849: Download page fetches all releases
**Bug Description**
The download page fetches all releases, whereas we only care about the latest one. At least on this page.

**Steps to Reproduce**
1. Go to [https://thenewboston.com/download](https://thenewboston.com/download)
2. Look at the network request to `/releases`

**Expected behavior**
Only the latest release should be fetched. The solution right now seems to be to use pagination.

**Actual behavior**
All releases are fetched.

**Screenshots/Recordings**
image

**OS and Browser**
All

**Additional context**

**Account Number**
Your account number where the reward coins for this bug bounty will be sent (pending review and approval).

ffpetrovic:
Fixes #849


ffpetrovic:
This could be improved by introducing something like [https://www.npmjs.com/package/query-string](https://www.npmjs.com/package/query-string) (**2kb gzipped**) and defining the query parameters for this request with that instead of hardcoding the query parameter format, because if it's decided to add more variable parameters to this method/request, this approach will get messy.

BundlePhobia analysis: [https://bundlephobia.com/result?p=query-string@6.13.7](https://bundlephobia.com/result?p=query-string@6.13.7)

PR #857, opened 2020-11-16
Utilise react-helmet-async instead of react-helmet

### LINKED ISSUE #856: UNSAFE_componentWillMount console error
**Bug Description**

**Steps to Reproduce**
1. Run TNB in local dev
2. Open the website's console
3. Observe the console error

**Expected behavior**
No console error should appear.

**Actual behavior**
Console error as shown below in the screenshot.

**Screenshots/Recordings**
Screen Shot 2020-11-16 at 4 21 50 PM

**OS and Browser**
 - OS: Mac, Chrome or Safari

**Additional context**
I have looked through to find a solution, and it seems that there is a problem with using `react-helmet`. We should utilise `react-helmet-async` instead. More info here: https://github.com/nfl/react-helmet/issues/548

**Account Number**

jamessspanggg:
Fixes #856 

Solution taken from: nfl/react-helmet#548


thesanjeevsharma:
@jamessspanggg I checked your solution solves the issue but it also introduced another bug.

When I go to any route for the first time, the title changes. But after that no matter how many routes I change, the title doesn't change at all. It always shows the title of the first page I visited.

Just try navigating to different pages and notice title.

jamessspanggg:
> @jamessspanggg I checked your solution, it solves the issue but it also introduced another bug.
> 
> When I go to any route for the first time, the title changes. But after that no matter how many routes I change, the title doesn't change at all. It always shows the title of the first page I visited.
> 
> Just try navigating to different pages and notice title.

My bad, fixed it. I moved the Provider out to the App component.

PR #865, opened 2020-11-16
Fix console warnings

### LINKED ISSUE #863: Fix console warnings during development
**Bug Description**
I can see two console warnings coming during development.

**Steps to Reproduce**
1) Go to Assets Page for the general react key warning.
2) The second one is little tricky to reproduce but it happens when you move from Home to any other page.

**Screenshots/Recordings**
Screen Shot 2020-11-16 at 8 32 00 PM

jamessspanggg:
Fixes #863 

First Asset page console issue: due to the `<PageTitle>` component being in a map function instead of the return function

Second Home page console issue is due to the timeout not being cleared when unmounting component, causing the setStates within setTimeout function to be called even when unmounted.

Acc no: c54c04774efaae20746fd3f29cdd619fea512e119bc1082b863572b2e8844104

### DIFF
--- a/src/containers/Assets/index.tsx
+++ b/src/containers/Assets/index.tsx
@@ -25,27 +25,27 @@ const assets: Asset[] = [
 const Assets: FC = () => {
   const renderCard = (asset: Asset): ReactNode => {
     return (
-      <>
-        <PageTitle title="Assets" />
-        <div key={asset.subtext} className="Assets__card-wrapper">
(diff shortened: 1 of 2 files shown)


jamessspanggg:
To replicate the second issue, try to insert a console log as follows, when `hi` is printed out to the console, navigate to a different page and observe the error.
Screen Shot 2020-11-17 at 12 04 49 AM

Issue #872, opened 2020-11-16
New Contributor


kristykjlee:
Team: [New User Operations]

{
  "title": [New User Operations Lead],
  "isLead": [true],
  "payPerDay": 2800,
  "createdDate": "2020-07-08T02:14:59.307535Z",
  "contributor": {
    "contributorId":[82ad4b185c2ac04440c8f1c54854819ac2ea374255e8fecc54a6f28d4fcc4814],
    "displayName": [Martyn],
    "githubUsername": [Spotnag],
    "profileImage": [PROFILE IMAGE URL (Slack or GitHub URL)],
    "slackUsername": [Martyn]
  }
}

VincentHch:
I'll take this one

VincentHch:
# aaf42054f84c66586bf29d446188347ab910112cb60b775a4d6feadd5b9ae73f

lapstjup:
@VincentHch  - Make sure to make the conributorId  = current maximum contributor id + 1.

PR #873, opened 2020-11-16
Add new user

VincentHch:


VincentHch:
#872

lapstjup:
@VincentHch - Since you might be unfamiliar with the process of adding a new team in our codebase, I have pushed a commit to address the same. You can get familiar with our codebase to solve more issues 😄 .

Issue #878, opened 2020-11-16
Kotlin SDK New Contributor Martin Nowosad


IanArb:
Team: Kotlin SDK

{
  "title": Team Lead,
  "isLead": true,
  "payPerDay": 2800,
  "createdDate": "2020-07-08T02:14:59.307535Z",
  "contributor": {
    "contributorId": [CONTRIBUTOR ID],
    "displayName": Martin Nowosad
    "githubUsername": https://github.com/MrIceman,
    "profileImage": https://ca.slack-edge.com/T07PJD1FZ-U016SCACHM4-1a824cfe8763-512
    "slackUsername": Martin Nowosad
  }
}

sdhiaeddine:
I will take this issue

PR #885, opened 2020-11-17
Kotlin New Contributor added

hasham56:
Fixed: #878


Hristijan95:
`contributorId` should not be null, it should be +1 from the latest contributor(51).

Issue #877, opened 2020-11-16
Kotlin SDK new member (Peter Sommerhoff)


IanArb:
Team: Kotlin SDK

{
  "title": Software Developer,
  "isLead": false,
  "payPerDay": 2800,
  "createdDate": "2020-07-08T02:14:59.307535Z",
  "contributor": {
    "contributorId": [CONTRIBUTOR ID],
    "displayName": Peter Sommerhoff,
    "githubUsername": https://github.com/petersommerhoff,
    "profileImage": https://avatars2.githubusercontent.com/u/2805791?s=460&u=5fcb7424015f63ae2194171c5649af332ab59f05&v=4,
    "slackUsername": Peter Sommerhoff,
  }
}

nparoski:
I can take this one

nparoski:
Since @sdhiaeddine made PR before me I will close my PR #880

PR #886, opened 2020-11-17
Kotlin developer Added

hasham56:
Fixed: #877 

### DIFF
--- a/src/data/teams.json
+++ b/src/data/teams.json
@@ -449,18 +449,17 @@
         }
       },
       {
-        "title": "Team Lead",
+        "title": "Software Developer",
         "isLead": true,
         "payPerDay": 2800,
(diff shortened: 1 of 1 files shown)


Hristijan95:
contributorId should not be null, it should be +1 from the latest contributor(51).
Also, the `isLead` field is incorrect.

hasham56:
@Hristijan95 Sir what if someone else do 52 for contributorId before me? what i did here is i added 52 for #878 and 53 for #877 ... is it ok? or should i recreate merge request?

Hristijan95:
If there is a conflict in some situation, I'll fix it before merging.

hasham56:
@Hristijan95 thanks ❤️ someone did fixed contributors id before merging my PR 😅

PR #887, opened 2020-11-17
Revert "Fix mobile drop-down menu blocking logo"

Hristijan95:


Hristijan95:
I had to revert this code because we discovered a bug in it.

PR #910, opened 2020-11-19
issue#909 - Long title words break the mobile design in the tasks page

mersimoski:
The titles that were big without spaces pushed the prices out of view, so i added a word break to keep mobile responsiveness.

From this:
image

To this:
image

Acc nr:
cf59a784ff69e64e977359f28cd4d04c0aea7a37cf1b8bda3dfbde2364032855

### DIFF
--- a/src/components/TopNav/TopNavPopoverButton/TopNavPopoverButton.scss
+++ b/src/components/TopNav/TopNavPopoverButton/TopNavPopoverButton.scss
@@ -7,7 +7,7 @@
   display: flex;
   font-weight: 700;
   justify-content: center;
-  outline: none;
+  outline: none!important;
   transition: color 0.1s ease;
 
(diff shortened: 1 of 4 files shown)


Hristijan95:
@mersimoski , we don't wanna use `!important` as stated here: https://thenewboston.com/style-guide/css
Also, please reference this PR to the issue.

mersimoski:
for this issue its just added word-break nothing else

lapstjup:
@mersimoski Use the word-wrap mixin available in **styles/_mixins.scss**. Also one PR should address one issue. You can raise separate PR for language addition and anything else which isn't related to this particular issue.

mersimoski:
@lapstjup i was just searching on how to send just PR for one issue and not to send every commit i have made.

Issue #914, opened 2020-11-19
New Contributor


colfax23:
Team: "Blockchain Research"

{
  "title": Blockchain Researcher,
  "isLead": false,
  "payPerDay": 2800,
  "createdDate": "2020-11-19T02:14:59.307535Z",
  "contributor": {
    "contributorId": ,
    "displayName": "Hargobind Gupta",
    "githubUsername": "Yara1990",
    "profileImage": https://ca.slack-edge.com/T07PJD1FZ-U01E9PS8ZEF-0edd4a970eba-512,
    "slackUsername": Hargobind -Blockchain Researcher
  }
}

hasham56:
Since @DinShpati added 2 contributors in 1 PR... So here is my alternative solution

PR #919, opened 2020-11-19
Alternative Solution to New Contributor

hasham56:
Fixed: #914 

### DIFF
--- a/src/data/teams.json
+++ b/src/data/teams.json
@@ -712,6 +712,24 @@
     ]
   },
   {
+    "title": "Research",
+    "contributors": [
+      {
+        "title": "Blockchain Researcher",
+        "isLead": false,
+        "payPerDay": 2800,
+        "createdDate": "2020-11-19T02:14:59.307535Z",
+        "contributor": {
+          "contributorId": null,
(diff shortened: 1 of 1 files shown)


src/data/teams.json
```diff
@@ -711,6 +711,24 @@
       }
     ]
   },
+  {
+    "title": "Research",
```
thesanjeevsharma, line 715: There's already a section for Research. Don't create a new one.

src/data/teams.json
```diff
@@ -711,6 +711,24 @@
...
+    "contributors": [
+      {
+        "title": "Blockchain Researcher",
+        "isLead": false,
+        "payPerDay": 2800,
+        "createdDate": "2020-11-19T02:14:59.307535Z",
+        "contributor": {
+          "contributorId": null,
```
thesanjeevsharma, line 723: Please add a `contributorId` here. Check the `contributorId`s in the development branch and choose the next in order.

For ex - if the highest `contributorId` right now is 55, use 56 here.

thesanjeevsharma (changes requested):

Issue #915, opened 2020-11-19
New Contributor


colfax23:
Team: "Blockchain Research"

{
  "title": Blockchain Researcher,
  "isLead": false,
  "payPerDay": 2800,
  "createdDate": "2020-11-19T02:14:59.307535Z",
  "contributor": {
    "contributorId": ,
    "displayName": "Kayode Babarinde",
    "githubUsername": "",
    "profileImage": "",
    "slackUsername": "Kayode Babarinde"
  }
}

hasham56:
Since @DinShpati added 2 contributors in 1 PR... So here is my alternative solution

PR #920, opened 2020-11-19
Alternative solution 2

hasham56:
Fixed: #915 


thesanjeevsharma (changes requested):
Same changes as requested on #919

PR #938, opened 2020-11-20
Fix NaN reward

### LINKED ISSUE #936: NaN reward
**Bug Description**
There seems to be NaN reward when running TNB website locally.

**Steps to Reproduce**
1. Run TNB website locally
2. Visit tasks
3. View that some tasks have NaN reward

**Expected behavior**
No NaN reward

**Actual behavior**
Some tasks have NaN reward

**Screenshots/Recordings**
Screen Shot 2020-11-20 at 9 46 16 PM

**OS and Browser**
 - OS: MacOS chrome

**Additional context**
Might be due to the new labels, causing to parsing of reward value to screw up

**Account Number**

jamessspanggg:
Fixes #936

Currently, the value of the reward label is directly parsed out using `parseInt`. With the introduction of other words such as `PR Reward 5000`, parseInt will fail and return NaN.
This PR handles the NaN due to parseInt by extracting only the numerical value in reward label.

Fixed screenshot:
Screen Shot 2020-11-20 at 10 13 36 PM

### DIFF
--- a/src/constants/regex.ts
+++ b/src/constants/regex.ts
@@ -1,0 +1,1 @@
+export const REGEX = {excludingNumber: /[^\d.]/g};
--- a/src/utils/github.ts
+++ b/src/utils/github.ts
@@ -2,6 +2,7 @@ import axios from 'axios';
 
 import {AMOUNT_COLOR, REPOSITORIES} from 'constants/github';
 import {BaseRelease, Issue, Release} from 'types/github';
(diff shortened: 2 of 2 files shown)


jamessspanggg:
I added the check for `.` within the regex as I was unsure if there will be decimal places in the future, i.e. `5000.00`. Do let me know if that's unnecessary and I'll remove it.

Issue #869, opened 2020-11-16
Make an Image component with optimizations for image loading


jamessspanggg:
**Current**
The leaderboard page loads github user avatars of size 460px, whereas the size of the div is 178px (desktop) and 54px (others). It is not needed to load such a big sized image for a small div.

May cause performance issues if there are more contributors, or the client does not have good device network.

Desktop:
<img width="1179" alt="Screen Shot 2020-11-17 at 12 40 02 AM" src="https://user-images.githubusercontent.com/32864116/99281554-9a49b480-286d-11eb-8954-b029bf5bcab4.png">

Mobile:
<img width="1430" alt="Screen Shot 2020-11-17 at 12 39 45 AM" src="https://user-images.githubusercontent.com/32864116/99281540-95850080-286d-11eb-9918-c69bad505acc.png">

**Suggestion**
Make an Image component and possibly optimise image loading using `srcset`, loading the appropriate size based on device widths.

Acc no: c54c04774efaae20746fd3f29cdd619fea512e119bc1082b863572b2e8844104

lapstjup:
@thesanjeevsharma @Hristijan95 I like this suggestion. Maybe we can have an Image component that abstracts all this and is reusable ?

Hristijan95:
I agree, would be nice to have that.

lapstjup:
@jamessspanggg I have increased the scope of this issue. Nice suggestion btw.

jamessspanggg:
> @jamessspanggg I have increased the scope of this issue. Nice suggestion btw.

May I clarify that the image component should be specific for the github avatar images? If so, then should the size of image at desktop and others be hardcoded or based on the passed in values?

lapstjup:
@jamessspanggg A reusable component would be good. Not specific to github and passed in values.

jamessspanggg:
Am busy with school exams atm, this issue is opened for anyone to solve!

PR #942, opened 2020-11-20
Added Avatar component

thesanjeevsharma:
fixes: #869

### DIFF
--- a/src/components/Avatar/Avatar.scss
+++ b/src/components/Avatar/Avatar.scss
@@ -1,0 +1,3 @@
+.Avatar {
+  border-radius: 50%;
+}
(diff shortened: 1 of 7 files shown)


Hristijan95 (approved):

lapstjup:
@thesanjeevsharma   - I made a change to prevent making network requests when we change window size for responsive tests. But there are more observations:-
1) Both the images are being loaded even though we use only either of mentioned sizes. 
2) When you switch from 7d to 30d view  - requests are being made each time and also the avatars are not rendered instantaneously compared to current implementation.

thesanjeevsharma:
@lapstjup 
image (1)

1. I checked network logs. Only one size is requested.
2.  Request will only be made for users that you did not see previously. This is because of lazy loading. For example, we have Kristy and me on 30d view, when you switch to All only one request is made - for avgust13. Kristy and my avatars are not fetched again.

PR #979, opened 2020-11-22
Added Hello World in Wolfo Language

dhakal0kushal:
#351 
Account: 22d0f0047b572a6acb6615f7aae646b0b96ddc58bfd54ed2775f885baeba3d6a


shahraizali:
language is not present in google translate but [ISO](https://www.loc.gov/standards/iso639-2/php/code_list.php) looks good. and the word itself is very close to Arabic/Persian which I somewhat understand so it appears to be correct. so approved from my end.

thesanjeevsharma:
@odetolataiwo 500

Issue #955, opened 2020-11-22
Update Docs for Bank - Config


buckyroberts:
Update documentation and sample code for Bank - Config
- ensure all available endpoints (`GET`, `POST`, `PATCH`, and `DELETE`) are documented
- update request and response examples from real API data (2 items for `GET` list responses)
- JSON should be formatted with 2 space indents
- ensure any/all params are documented in tables
- https://thenewboston.com/bank-api/config

velibor7:
I will take this

VincentHch:
@buckyroberts you can close this issue, the documentation for all config endpoints are up-to-date

velibor7:
@VincentHch Not exactly, for example type of default_transaction_fee is integer on API, but in docs its a string, I just fixed it I think there is 2 or 3 more lines, also small mistakes, but not up to date :)

shahraizali:
@velibor7 Good catch bud 👍

Issue #951, opened 2020-11-22
Update Docs for Bank - Accounts


buckyroberts:
Update documentation and sample code for Bank - Accounts
- ensure all available endpoints (`GET`, `POST`, `PATCH`, and `DELETE`) are documented
- update request and response examples from real API data (2 items for `GET` list responses)
- JSON should be formatted with 2 space indents
- ensure any/all params are documented in tables
- https://thenewboston.com/bank-api/accounts

velibor7:
I will take this and do it later today :)

PR #982, opened 2020-11-22
Documentation fixes

velibor7:
accounts PATCH is not allowed according to the API so I commented it.
blocks POST was already okay


velibor7:
This is fixing issues: #951 #954 #955

shahraizali:
@velibor7 were you able to test the POST and other methods with a local setup?

src/containers/BankApi/BankApiAccounts/index.tsx
```diff
@@ -18,25 +18,31 @@ const BankApiAccounts: FC = () => {
 
       <DocEndpoint endpoint="/accounts" method="GET" />
       <RequestResponseSnippet
```
shahraizali, line 20: Should we add query parameters as well here? like @VincentHch added [here](https://github.com/thenewboston-developers/Website/pull/980/files)

velibor7: query params fixed with 9e1a895
I got signature error since I'm not sure how to generate valid one

Also, important questions:
Should we set count to 2 or to the value provided on the API?
And should I make my result as Vincent did {}, {}, ... {} or should we do it this way (with just 2 elements as I did)

shahraizali: Your way is correct. We just need 2 records as mentioned in the descriptions. but that would mean that the next will have value like limit=2 but I don't mind it being anything since we just want to share the structure here. So, you are good in my review. let's see if the community has some other opinion on it 👍

shahraizali: For the POST APIs, we need to set up projects locally or what we can do is set up a public test one for devs and share the signature keys so not everyone has to go through the hassle to setup. Thoughts? @velibor7 @VincentHch @sbalasa @DucPhamTV @Dragan @makaramkd

VincentHch: To verify the POST request (for `/invalid_blocks`), I looked directly at the code.
Here is the invalid block sent by the validator to the bank : https://github.com/thenewboston-developers/Validator/blob/master/v1/tasks/confirmation_block_queue.py#L128

makaramkd: Maybe for now we should stick with setting up the project locally as I can see we are the only one working on it.

DucPhamTV: Of course setting up a local env to test is great but I'm still not able to setup successfully. Currently I'm using my bank which is joined the network to do POST/PATCH APIs.

shahraizali: Okay, let's setup our nodes locally first. I see a channel already created named 'node-help' we can discuss there if someone hits a bump so others who want to set up can get help as well.

src/containers/BankApi/BankApiAccounts/index.tsx
```diff
@@ -18,25 +18,31 @@ const BankApiAccounts: FC = () => {
...
-    "id": "ae4d43b0-5c34-4e56-8266-0e3531268815",
-    "created_date": "2020-07-08T02:15:12.271834Z",
-    "modified_date": "2020-07-08T02:15:12.271852Z",
-    "account_number": "a29baa6ba36f6db707f8f8dacfa82d5e8a28fa616e8cc96cf6d7790f551d79f2",
-    "trust": "94.63"
-  }
-]`}
+        code={`{
+  "count": 2,
+  "next": "http://143.110.137.54/accounts?limit=50&offset=50",
```
shahraizali, line 23: add params table for this api as well and the ideal response to add I believe should be the response of the following url.
http://143.110.137.54/accounts?limit=2

velibor7: Fixed it for accs and for blocks too, to make it consistent

src/containers/BankApi/BankApiAccounts/index.tsx
```diff
@@ -17,26 +17,47 @@ const BankApiAccounts: FC = () => {
...
+            "trust": "0.00"
+        }
+    ]
+}`}
         heading="Response"
       />
 
+      {/** 
+       
       <DocEndpoint endpoint="/accounts/<account_number>" method="PATCH" />
```
shahraizali, line 61: why is this commented is this endpoint not in the code?

velibor7: The code was there before, but according to the api it is not allowed, so I commented it. Should I delete it?

shahraizali: looking at the code looks like patch is allowed. There is **UpdateModelMixin** and **update** function inside the class which means it accepts both PUT and PATCH also notice the argument **partial=True**.

shahraizali (approved):

PR #983, opened 2020-11-22
Updated Docs for Primary Validator - Accounts #962 Fix

### LINKED ISSUE #962: Update Docs for Primary Validator - Accounts
Update documentation and sample code for Primary Validator - Accounts
- https://thenewboston.com/primary-validator-api/accounts

Hristijan95:
Fixes #962.
@buckyroberts, for the URL params (/accounts/<account_number>/balance for example) I think we are not adding them into tables, correct? If we are, I'll add a table, but as far as I could see, we only use tables for body params. If I am wrong, just let me know, I'll add the tables

### DIFF
--- a/src/containers/PrimaryValidatorApi/PrimaryValidatorApiAccounts/index.tsx
+++ b/src/containers/PrimaryValidatorApi/PrimaryValidatorApiAccounts/index.tsx
@@ -52,7 +52,7 @@ const PrimaryValidatorApiAccounts: FC = () => {
       <TableParams
         items={[
           {
-            dataType: 'URL parameter',
+            dataType: 'PATH parameter',
             description: 'The account number that you want to check',
             param: 'account_number',
           },
@@ -69,7 +69,7 @@ const PrimaryValidatorApiAccounts: FC = () => {
       <TableParams
         items={[
           {
-            dataType: 'URL parameter',
+            dataType: 'PATH parameter',
             description: 'The account number that you want to check',
             param: 'account_number',
           },
--- a/src/containers/PrimaryValidatorApi/PrimaryValidatorApiAccounts/index.tsx
+++ b/src/containers/PrimaryValidatorApi/PrimaryValidatorApiAccounts/index.tsx
@@ -11,6 +11,20 @@ const PrimaryValidatorApiAccounts: FC = () => {
       </p>
 
       <DocEndpoint endpoint="/accounts" method="GET" />
+      <TableParams
+        items={[
+          {
+            dataType: 'URL parameter',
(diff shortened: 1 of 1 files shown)


shahraizali:
Hi @Hristijan95! I confirmed with bucky and we did a small poll on the documentation channel as well we would be adding the query params to the tables as well. Look [here](https://github.com/thenewboston-developers/Website/pull/982/files#diff-c4eee41d3a8c9ab0d7ec0cb4717b0991755b71e565eb72fe46e6217c8b5be168) for reference

Hristijan95:
Okay, great, I added the tables.

src/containers/PrimaryValidatorApi/PrimaryValidatorApiAccounts/index.tsx
```diff
@@ -12,44 +12,55 @@ const PrimaryValidatorApiAccounts: FC = () => {
...
-    "balance_lock": "5e12967707909e62b2bb2036c209085a784fabbc3deccefee70052b6181c8ed8"
-  },
-  {
-    "id": "e8b24ad7-f81e-4731-818e-9bbbc3336428",
-    "account_number": "ad1f8845c6a1abb6011a2a434a079a087c460657aad54329a84b406dce8bf314",
-    "balance": "56.0000000000000000",
-    "balance_lock": "ad1f8845c6a1abb6011a2a434a079a087c460657aad54329a84b406dce8bf314"
-  }
-]`}
+        code={`{
```
shahraizali, line 15: we are adding Table for Query params as well like **limit** and **offset**. kindly include that as well
Have a look here
https://github.com/thenewboston-developers/Website/pull/982/files#diff-c4eee41d3a8c9ab0d7ec0cb4717b0991755b71e565eb72fe46e6217c8b5be168

Hristijan95:
@shahraizali I added the pagination query params, but I think we should make a difference between the query params & URL parameters, or I should remove the url/path parameters? 

To be more specific, I am talking about differentiating `/accounts/<account_number>/balance` and `/accounts?limit=50&offset=50`
Right now both are called "URL parameter". 
maybe account_number could stay "URL parameter" and the query params should be "Query parameter"? Let me know what you think.

shahraizali:
How about renaming the one that is part of url to 'Path parameters' 
SOURCE: https://idratherbewriting.com/learnapidoc/docapis_doc_parameters.html

as far as the name URL parameter and Query Parameter I also looked it up a couple of days ago when I saw your PR. In my finding, they are used interchangeably and are arguably the same thing, In my opinion, that's fine.

Hristijan95:
Agreed. I changed them to PATH.

shahraizali (approved):

PR #987, opened 2020-11-22
Hello world Swedish correction

Godnoken:
"Hello World" should be "Hallå Världen" or at the very least "Hej Världen".

"Hej Värld" is just incorrect and looks like an automatic translation.

In case this fix is rewarded:

Account ID; ed0026cc11eef78c454107ed1d3ad3b1d3a4f1628ea2743e7eb4e1daf0a95b2e


Hristijan95:

PR #981, opened 2020-11-22
[Documentation] Bank invalid blocks

VincentHch:
#957

PR #989, opened 2020-11-23
964-docs primary validator banks

### LINKED ISSUE #964: Update Docs for Primary Validator - Banks
Update documentation and sample code for Primary Validator - Banks
- https://thenewboston.com/primary-validator-api/banks

aayushi14:
closes #964


src/containers/PrimaryValidatorApi/PrimaryValidatorApiBanks/index.tsx
```diff
@@ -9,19 +9,35 @@ const PrimaryValidatorApiBanks: FC = () => {
...
-    "port": 8000,
-    "protocol": "http",
-    "version": "v1.0",
-    "default_transaction_fee": "1.0000000000000000",
-    "confirmation_expiration": null,
-    "trust": "0.00"
-  }
-]`}
+        code={`{
+    "count": 20,
```
aayushi14, line 13: Does the count value need to be changed to "count":2, as we are showing 2 results in snippet?

shahraizali: Hi! no, the count is the total number of records. Ideally, we would want to show the user the next object as well so if you hit the URL like this
http://143.110.137.54/banks?limit=2
it will give 2 records and we will have the next object as well so the user can know how to fetch more records if required since we are just showing 2. 

Also, some PRs have Param Table added for example [here](https://github.com/thenewboston-developers/Website/pull/981/files) but it's not yet final whether we are going to include the Query Params in the docs or not.

shahraizali: Hi! no, the count is the total number of records. Ideally, we would want to show the user the next object as well so if you hit the URL like this
http://143.110.137.54/banks?limit=2
it will give 2 records and we will have the next object as well so the user can know how to fetch more records if required since we are just showing 2. 

Also, some PRs have Param Table added for example [here](https://github.com/thenewboston-developers/Website/pull/981/files) but it's not yet final whether we are going to include the Query Params in the docs or not.

PR #1001, opened 2020-11-23
977-updated docs cv - validators

### LINKED ISSUE #977: Update Docs for Confirmation Validator - Validators
Update documentation and sample code for Confirmation Validator - Validators
- https://thenewboston.com/confirmation-validator-api/validators

aayushi14:
closes #977


src/containers/ConfirmationValidatorApi/ConfirmationValidatorApiValidators/index.tsx
```diff
@@ -9,36 +9,41 @@ const ConfirmationValidatorApiValidators: FC = () => {
...
-    "version": "v1.0",
-    "default_transaction_fee": "4.0000000000000000",
-    "root_account_file": "https://gist.githubusercontent.com/buckyroberts/519b5cb82a0a5b5d4ae8a2175b722520/raw/9237deb449e27cab93cb89ea3346ecdfc61fe9ea/0.json",
-    "root_account_file_hash": "4694e1ee1dcfd8ee5f989e59ae40a9f751812bf5ca52aca2766b322c4060672b",
-    "seed_block_identifier": "",
-    "daily_confirmation_rate": null,
-    "trust": "100.00"
-  }
-]`}
+        code={`{
```
shahraizali, line 12: we have decided to add the table of query params kindy add that.
[here](https://github.com/thenewboston-developers/Website/pull/982/files#diff-c4eee41d3a8c9ab0d7ec0cb4717b0991755b71e565eb72fe46e6217c8b5be168) is the reference of how that should look like

aayushi14: Also added GET: /validators/<node_identifier>, can you check and let me know if this is correct?

src/containers/ConfirmationValidatorApi/ConfirmationValidatorApiValidators/index.tsx
```diff
@@ -1,44 +1,86 @@
...
+            "daily_confirmation_rate": null,
+            "trust": "100.00"
+        }
+    ]
+}`}
+        heading="Response"
+      />
+
+      <DocEndpoint endpoint="/validators/<node_identifier>" method="GET" />
+      <TableParams
```
shahraizali, line 60: Param in the table I think is not correct where did you pick it from? The response is correct.

src/containers/ConfirmationValidatorApi/ConfirmationValidatorApiValidators/index.tsx
shahraizali: where did you pick these TableParams from? They seem incorrect here params means the one we send in the url like **limit**
http://54.219.178.46/validators?limit=1
specifying this param will limit the results to 1. but you mentioned trust which is not an added filter
so doing something like http://54.219.178.46/validators?trust=1
does not do anything for me.

src/containers/ConfirmationValidatorApi/ConfirmationValidatorApiValidators/index.tsx
aayushi14: image

shahraizali: What you highlighted means that when you send a PATCH request or use Update method then the parameter to send along would be trust. so you can update the trust value of validator with a Patch request but the description of list is just 1 line below the list then the update description starts.

src/containers/ConfirmationValidatorApi/ConfirmationValidatorApiValidators/index.tsx
aayushi14: Oh okay got it. Thanks for explaining

shahraizali (approved):

PR #1026, opened 2020-11-25
Update documentation for Connection Requests

DucPhamTV:
I think #974 , #967 , #958 are the same issues.

### DIFF
--- a/src/containers/NodeApi/NodeApiConnectionRequests/index.tsx
+++ b/src/containers/NodeApi/NodeApiConnectionRequests/index.tsx
@@ -1,6 +1,6 @@
 import React, {FC} from 'react';
 
-import {DocContainer, DocEndpoint, DocImage, RequestResponseSnippet} from 'components';
+import {DocContainer, DocEndpoint, DocImage, RequestResponseSnippet, TableParams} from 'components';
 
 import ConnectionRequestDiagram from './ConnectionRequest.png';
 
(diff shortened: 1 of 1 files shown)


shahraizali:
Hey, @DucPhamTV  sorry for the delayed review. They seem to be the same but are different in a way that they are different nodes like when they are running all of these three would have different IPs and the URL might be similar but the response can be different. Like for example **/clean and /crawl** are the same URL for primary and confirmation validators but works for confirmation and not for primary similarly kindly verify these on your end as well and you can post your findings here also update the docs accordingly if required.

DucPhamTV:
Hi @shahraizali , I have tested /connection_requests to 3 types of node. They are the same request and response format.

- Between Bank and CV, I can send /connection_requests outside the node to the target node.
- PV won't send /connection_requests. It receives /connection_requests from Bank and CV. I tried to send the request outside the Bank or CV, It doesn't work since there are some additional steps before connection_requests is sent. So I added new code to dump the request and response.

shahraizali:
> They are the same request and response format

Still, I think we should update all three endpoints in docs since these are three different concepts and urls documentation wise.

https://thenewboston.com/confirmation-validator-api/connection-requests
https://thenewboston.com/primary-validator-api/connection-requests
https://thenewboston.com/bank-api/connection-requests

DucPhamTV:
If we separate it to 3 different endpoints. I will:
- Create 3 new Images to demonstrate the connections
- Move common NodeApiConnectionRequests to 3 endpoints folders
- Update text `index.tsx` according to new location.
What do you think?

buckyroberts:
Fixes https://github.com/thenewboston-developers/Website/issues/974 https://github.com/thenewboston-developers/Website/issues/967 https://github.com/thenewboston-developers/Website/issues/958

PR #1070, opened 2020-12-02
Add Profile Info section

### LINKED ISSUE #1062: Create the "Profile Info" section of the profile page
This issue is part of #315 , the idea is to create the left (profile info) section of the page. You can check the Figma design here: https://www.figma.com/file/ChyT88EmP2Rmzx35I9erCL/Marketing-Site?node-id=1353%3A1915

Just to note, the avatar has a white stroke (2px) around it, so it doesn't blend in with the background color.
On the 1366px screens, the user info block & the task completed block that's been designed is the max width of those containers, for smaller screens, the blocks will respond according to the tablet/mobile

There's an existing empty component in `src/containers/Profile/index.tsx` . This is the component where the "Profile Info" section and the "Tasks Completes" section should go. Use this section to place your component that you should create inside the `src/containers/Profile` directory.

jamessspanggg:
Fixes #1062 

Preview of addition: http://tnb-profile-info.surge.sh/user/profile


jamessspanggg:
Before I continue further, there are some doubts:

1. Is the profile openly accessible to anyone? If so, the edit cover image button should not be visible if the authenticated user is not the owner of the profile?
2. The `teams` and `titles` within the `TeamMember` interface is currently not used, as the JSON files do not include these fields. Will these fields be populated in the future or are they not gonna be used? (This is important for finding all the positions of the team member)
3. May I ask what will be passed to the Profile component? I assume that it is the authenticated user, but for now can I just pass in the github_username and get his/her details based on that?
4. Also, a user may not be a full time team member, which do not have position titles, slack username etc. For this case, should I just omit the UI that shows these?

@Hristijan95 do u happen to know about the solutions to these questions?

lapstjup:
@jamessspanggg Brilliant work 🚀 . 

_May I ask what will be passed to the Profile component? I assume that it is the authenticated user, but for now can I just pass in the github_username and get his/her details based on that?_
**This is yet to be decided. This page is exclusive to logged in users only. I saw your implementation. For tasks as well as profile info, you're fetching all data from github using `github_username` as prop. But most probably what will happen is you'll get data from redux store which you will pass as props to these components and top level routing will ensure to only show this page for logged in users. For now this is cool.** 

_Also, a user may not be a full time team member, which do not have position titles, slack username etc. For this case, should I just omit the UI that shows these?_
**Yeah these can be emitted in those scenarios.**

_In the design, it seems like there will be links for teams, i.e. Front-End Developers. I assume that this links to the /teams page. But currently, under the teams page, there's no specific endpoint for each team (i.e. /teams/frontenddevs). What link should I place for these?_
**You can skip this for now**

lapstjup:
@thesanjeevsharma and @Hristijan95  - In terms of design, this is ready. But I think it's good to only merge this post redux impl. What do you think ?

thesanjeevsharma:
@lapstjup Yes! Let's hold it over weekend at least. After the call tomorrow, we'll have a better clarity and even @Hristijan95 would have made some progess with redux.

Hristijan95:
@lapstjup, @thesanjeevsharma, we'll have to merge it in any case before the API/Redux implementation because I'll need the code to wire everything together.

One small issue though, @jamessspanggg, the account number is almost collided with the QR Code on my laptop:
Screenshot 2020-12-07 at 16 31 20

jamessspanggg:
> One small issue though, @jamessspanggg, the account number is almost collided with the QR Code on my laptop:
> Screenshot 2020-12-07 at 16 31 20

Hmm I can't reproduce it though, may I ask what browser and laptop u are using?

jamessspanggg:
@Hristijan95 realised what was the issue. I assume you were using Safari? The css property `gap` for flex displays are not available for Safari. I have fixed the issue.

src/containers/Profile/ProfileInfo/index.tsx
```diff
@@ -49,11 +49,11 @@ const ProfileInfo: FC<ComponentProps> = ({github_username}) => {
             </div>
           ))}
         <div className="ProfileInfo__member-slack">
-          <Icon icon={IconType.slack} onClick={() => null} size={18} />
+          <Icon className="ProfileInfo__member-slack-icon" icon={IconType.slack} onClick={() => null} size={18} />
```
lapstjup, line 52: Hey I wanted to know the reason to pass `()=>null` to `onClick` here ?

jamessspanggg: Opps my bad forgot to remove this

lapstjup:
@jamessspanggg Can you confirm on the behaviour of opening edit modal in mobile view. I tried it in my iPhone and the modal's header isn't visible until I scroll below.

jamessspanggg:
> @jamessspanggg Can you confirm on the behaviour of opening edit modal in mobile view. I tried it in my iPhone and the modal's header is partially visible until I scroll below.

Yes can confirm that Iphone safari it is partially visible. Might be due to the previous implementation of Modal not respecting the mobile browser's top url bar. Let me get back to this tomorrow, it's quite late at night here

src/components/Layout/index.tsx
```diff
@@ -22,7 +23,9 @@ const Layout: FC<ComponentProps> = ({children}) => {
       <div className="Layout__top-nav-wrapper">
         <TopNav className="Layout__TopNav" />
       </div>
-      <div className={clsx({Layout__content: !isHomepage, Layout__home: isHomepage})}>{children}</div>
+      <div className={clsx({Layout__content: !(isHomepage || isProfilePage), Layout__home: isHomepage})}>
```
jamessspanggg, line 26: @lapstjup may I ask why is this change needed? after your modifications on this, the profile page no longer respects the 1280px max width and its width is taking the entire page

lapstjup: Yeah so I did this to be consistent with Figma design. The light blue background is across the whole width. Does this page need to respect that 1280px max width for any use case ?

jamessspanggg: Oh right, to me it looks slightly weird when on larger screens, also it stays consistent with the leaderboard page (also contains tasks completed) which also has max width of 1280px

lapstjup: Let's keep it for now for consistency sake.

jamessspanggg: tested and works fine. thanks for the help

lapstjup: No worries. Thanks a ton for confirming. Now go sleep pretty late man 😴 .

PR #1097, opened 2020-12-08
Github Authentication

pgeorge2:
- Changed the Backend API Url logic to make login with github feature works


buckyroberts:
Backend for this: https://github.com/thenewboston-developers/Website-API/pull/29

.env.example
```diff
@@ -1,2 +1,2 @@
-REACT_APP_BACKEND_API=
+REACT_APP_BACKEND_API=http://127.0.0.1:8000/accounts/github/login/
```
jamessspanggg, line 1: should these be added to the `.env` file instead? seems like these are not loaded to `process.env`

pgeorge2: Yes. This should be added into .env file. I added this in .env.example for reference.

jamessspanggg (approved):
LGTM, can wait for the front end leads for further review

src/components/TopNav/index.tsx
```diff
@@ -171,12 +171,12 @@ const TopNav: FC<ComponentProps> = ({className}) => {
         <Link className={clsx('TopNav__right-item', 'TopNav__download-button')} tabIndex={-1} to="/download">
           <Button>Download</Button>
         </Link>
-        {/* <A
+        <A
           className={clsx('TopNav__right-item', 'TopNav__download-button')}
-          href={`https://github.com/login/oauth/authorize?client_id=${process.env.REACT_APP_GITHUB_CLIENT_ID}&scope=read%3auser%20user%3aemail%20`}
+          href={`${process.env.REACT_APP_BACKEND_API}`}
```
lapstjup, line 176: This particular line will remain same I think. Because the user will route to the github login page where the sign in will happen and the user will get redirected to a callback url having `code` in query params which will be sent to `REACT_APP_BACKEND_API` so that further backend can use that code to retrieve access token. 
I hope I am right :p

lapstjup: Also let's keep this commented until the whole flow is done.

vosi: not quite right, we are doing auth on a backend side, so, the flow is to go to our URL, that will redirect to a proper github URI

vosi (approved):

src/components/TopNav/index.tsx
lapstjup: @vosi So if understood this right, this is the flow:-
1) User clicks on login and request for login goes to backend. 
2) Backend redirects the user to the Github sign in URL, if user is not already signed in.
3) User sign's in and is redirected to our website's callback URL with `code` which we send back to backend.

buckyroberts:
Sticking a "Don't Merge" label on it to make sure this doesn't get merged in without the backend https://github.com/thenewboston-developers/Website-API/pull/29

vosi:
to whom it may concern      

here is an example of the frontend authentication flow using some pseudocode (aka jQuery)
https://github.com/thenewboston-developers/Website-API/blob/3a4f01939184b7dc242942a3e22f4c1f21a0c2b1/v1/testapp/templates/home.html

it is not merged yet, there are some small problems caused by a custom User model, the thing does not acquire a user's name from GitHub properly

but I believe one can start implementing a proper code in React

buckyroberts:
Closing in place of https://docs.google.com/document/d/1icup7KaI7uJd2ryjCVl7V8OjBPKFVIngw7xXlIbdXeI/edit?usp=sharing

PR #1118, opened 2020-12-14
Prevent mobile menu from blocking header elements during animation

SirDaev:
Account: e5c2ff3c4f6886fd7cd87002d83c6aabe48e585ce38026c0df6ea7c517e0f7a4

### DIFF
--- a/src/components/TopNav/TopNav.scss
+++ b/src/components/TopNav/TopNav.scss
@@ -48,7 +48,7 @@ $top-nav-height: 60px;
     background-color: var(--color-bg);
     display: flex;
     flex: 1;
-    height: 60px;
+    height: $top-nav-height;
     position: relative;
     z-index: 3;
   }
--- a/src/components/TopNav/TopNav.scss
+++ b/src/components/TopNav/TopNav.scss
@@ -43,6 +43,16 @@ $top-nav-height: 60px;
     }
   }
 
+  &__left {
+    position: relative;
+    flex: 1;
+    background-color: var(--color-bg);
+    z-index: 3;
+    height: 60px;
+    display: flex;
+    align-items: center;
+  }
+
   &__right {
     align-items: center;
     display: flex;
@@ -61,6 +71,7 @@ $top-nav-height: 60px;
 
     @media (max-width: 992px) {
       display: block;
+      background-color: var(--color-bg);
     }
 
     &__button {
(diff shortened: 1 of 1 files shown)


jamessspanggg:
@SirDaev the linting seems to be failing, I suspect it is due to the CSS linter violations. Can you run `npm run stylelint` locally and push again?

SirDaev:
@jamessspanggg Thanks, I missed that. Should be all set!

src/components/TopNav/TopNav.scss
```diff
@@ -43,6 +43,16 @@ $top-nav-height: 60px;
     }
   }
 
+  &__left {
+    align-items: center;
+    background-color: var(--color-bg);
+    display: flex;
+    flex: 1;
```
jamessspanggg, line 50: Do you need the `flex: 1` here? seem like not affecting anything

src/components/TopNav/TopNav.scss
```diff
@@ -43,6 +43,16 @@ $top-nav-height: 60px;
...
+    height: 60px;
```
jamessspanggg, line 51: can utilise `$top-nav-height` here instead of a magic number

jamessspanggg (commented):
Tested and indeed the mobile menu does not block the header elements. Here are some comments

src/components/TopNav/TopNav.scss
SirDaev: I added `flex: 1` so that the entire drop menu would be covered until it gets below the header. Currently the logo appears to hide the menu as the menu drops, but only because the text content of the current menu items is not as wide as the logo.

Example: (change 'Join the Community!' to 'Join the Community! It is amazing!')

With `flex: 1` the menu item is hidden until below `.TopNav__left`
Screen Shot 2020-12-15 at 8 33 19 AM

Without `flex: 1` the menu item will appear to the right of the logo on its way down.
Screen Shot 2020-12-15 at 8 34 37 AM

Technically this doesn't directly effect "blocking logo and icon", but I saw it as a future-oriented improvement.

src/components/TopNav/TopNav.scss
SirDaev: Good call! I'll make the change.

jamessspanggg (approved):
LGTM 👍🏻

PR #1122, opened 2020-12-14
Redux implementation

### LINKED ISSUE #1080: Redux implementation
This issue will be a part of #315. 
Make a similar redux-toolkit implementation like in the Account-Manager for the needs of the whole login/registration functionality.

I'll be working on it.

Hristijan95:
Fixes #1080


src/store/user/authentication.ts
```diff
@@ -0,0 +1,18 @@
...
+import {APP_AUTHENTICATION} from 'constants/redux';
+import {Dict} from 'types/store';
+
+const authentication = createSlice({
+  initialState: {} as Dict<any>,
+  name: APP_AUTHENTICATION,
+  reducers: {
+    setAuthentication: (state, {payload}: PayloadAction<any>) => {
+      const {accessToken} = payload;
+      state.accessToken = accessToken;
```
thesanjeevsharma, line 11: I think we should rename this to `token` or `authToken` as we are already using `accessToken` in backend for some other purpose(GitHub). Not a problem, but just to be clear.

Hristijan95: I just put `accessToken` for now because I am not sure what will we receive from the API once it's done. So basically it's just a placeholder for now, we'll change it when we implement authentication.

thesanjeevsharma (approved):

src/store/user/profile.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const profile = createSlice({
+  initialState: {} as Dict<any>,
+  name: USER_PROFILE,
+  reducers: {
+    setProfile: (state, {payload}: PayloadAction<any>) => {
+      const {name, account, githubUsername} = payload;
+      state.name = name;
+      state.account = account;
+      state.githubUsername = githubUsername;
```
lapstjup, line 13: So with these slices, its not required to create a copy of state and we can directly mutate it ?. Is there internally something handling this ?

thesanjeevsharma: yeah, immer does it.

src/dispatchers/user/index.ts
```diff
@@ -0,0 +1,9 @@
+import axios from 'axios';
+
+import {setAuthentication} from 'store/user';
+import {AppDispatch} from 'types/store';
+
+export const authenticateUser = () => async (dispatch: AppDispatch) => {
+  const {data} = await axios.post<any>(`${process.env.REACT_APP_BACKEND_API}/login`);
```
lapstjup, line 7: `authenticateUser` will take a parameter to post that to backend right ?

Hristijan95: Yep, sorry, for responding this late, I've missed the comment.
At the moment this was put as just an example until the API is done.

PR #1124, opened 2020-12-15
Link leaderboard and task to profile page

### LINKED ISSUE #1123: Link leaderboard and task page to profile page
- [ ] On the leaderboard, the users github names will now link to the users profile.
- [ ] On the tasks page, clicking on the users avatar will link to the users profile.

Part of #315

jamessspanggg:
Fixes #1123 

Tasks page: Clicking on the profile image of the assignee will be directed to their corresponding profiles
Leaderboard page: Clicking on the username will be directed to their corresponding profiles

### DIFF
--- a/src/containers/App.tsx
+++ b/src/containers/App.tsx
@@ -81,7 +81,7 @@ const App: FC = () => {
           <Route path="/project-proposals/:chapter?" component={ProjectProposals} />
           <Route path="/style-guide/:chapter?" component={StyleGuide} />
           <Route path="/user/signin/callback" component={Auth} />
-          <Route path="/user/profile" render={() => <Profile github_username="jamessspanggg" />} />
+          <Route path="/user/profile/:githubUsername" render={() => <Profile />} />
           <Redirect to="/" />
         </Switch>
       </Layout>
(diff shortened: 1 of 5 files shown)


angle943:
@jamessspanggg @Hristijan95 what is the status of this PR? Why did it get marked as `DONT MERGE`?

Hristijan95:
@angle943  if I remember correctly, it's marked like that because at the time the profile screen was non existent, or something like that. I am from my phone right now, I can check in more details in the morning.

jamessspanggg:
@angle943 since users can edit their profile's `github_account` easily, it will be difficult to achieve this as there might be multiple accounts with the same `github_account`.

angle943:
@jamessspanggg i think we should change it to `/user/profile/:uuid`, since people will have unique uuids

AbhayAysola:
Maybe we could use Github OAuth and make sure 2 people don't have the same Github username.

jamessspanggg:
> @jamessspanggg i think we should change it to `/user/profile/:uuid`, since people will have unique uuids

But from the leaderboard data, it is difficult to get the `uuid` currently, we will probably need to update the `contributors` to include their `uuid`.

angle943:
@jamessspanggg hmm yeah thats right. lets discuss this with @buckyroberts

Issue #1139, opened 2020-12-21
New Contributor


Rajputusman:
Team: [Discord Managers]

{
  "title": [Moderator],
  "isLead": [false],
  "payPerDay": 2800,
  "createdDate": "2020-07-08T02:14:59.307535Z",
  "contributor": {
    "contributorId": [CONTRIBUTOR ID],
    "displayName": [Muhammed Medo],
    "githubUsername": [codemedo],
    "profileImage": [https://avatars0.githubusercontent.com/u/48848156?v=4],
    "slackUsername": [Muhammed Medo]
  }
}

Rajputusman:
contributorId = 64

PR #1140, opened 2020-12-21
New Contributor

Rajputusman:
fixes: #1139

### DIFF
--- a/src/data/teams.json
+++ b/src/data/teams.json
@@ -336,10 +336,8 @@
           "profileImage": "https://avatars2.githubusercontent.com/u/25906602",
           "slackUsername": "DinShpati"
         }
-      }
-    ]
-  },
-  {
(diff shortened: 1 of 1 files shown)


jamessspanggg:
How did u decide that the contributor id is 64? isn't it 62 as the last contributor id is 61?

Rajputusman:
@jamessspanggg I got the contributor ID from the payment repo for teams.csv
image
image
image

jamessspanggg:
@Rajputusman I think the two need not to be match though, because for the user "Nish", he has a contributorId of 61 (in teams.json) instead of 63 (teams.csv). So to follow order of teams.json, we should use contributorId of 62?

jamessspanggg (approved):

thesanjeevsharma (approved):

PR #1100, opened 2020-12-09
Top nav changes

amarjeet987:
Navbar updates as per the new UI

PR #1151, opened 2020-12-27
Authentication, Profile, and Redux

buckyroberts:
Consolidating multiple branches into this one for easier development.

- https://github.com/thenewboston-developers/Website/pull/1100
- https://github.com/thenewboston-developers/Website/pull/1122


src/components/AuthContainer/AuthContainer.scss
```diff
@@ -2,8 +2,8 @@
   border-radius: 12px;
   box-shadow: 0 0 10px rgba(85, 108, 214, 0.18);
   margin: 48px auto 64px;
+  max-width: 480px;
```
jamessspanggg, line 5: Changed width -> max width as smaller devices have the following UI:

Screen Shot 2020-12-28 at 2 58 09 PM

buckyroberts:
To do tomorrow:
- [ ] Fix bottom spacing on mobile
- [ ] Add create account and sign up links on mobile menu

screencapture-localhost-3000-sign-in-2020-12-28-02_37_00

jamessspanggg:
@buckyroberts is this ready for review?

buckyroberts:
@jamessspanggg There are still some tweaks to make and the edit modal needs hooked up but you can start reviewing. To test locally you can create a file called `.env.local` and set the contents to:
```
REACT_APP_BACKEND_API=http://api-dev.thenewboston.com
```

You can sign in with a test user using:
```
dev@email.com
Apples123!
```

You can also test the new account creation but to complete the process I still need to manually verify the email on the backend (since our email system isn't hooked up yet)

src/containers/SignIn/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+    try {
+      const {data} = await dispatch(login({email, password}));
+      history.push(`/users/${data.pk}`);
+    } catch (error) {
+      setErrorMessage(formatAPIError(error));
+    }
+  };
+
+  const validationSchema = yup.object().shape({
+    email: yup.string().required('This field is required'),
```
jamessspanggg, line 34: should we use `yup.string().email()...` here?

src/containers/CreateAccount/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+        Our email verification system is still in development. Please contact Bucky on Slack to manually verify your
+        email.
+      </p>
+    );
+  };
+
+  const validationSchema = yup.object().shape({
+    confirmPassword: yup.string().oneOf([yup.ref('password'), ''], 'Passwords must match'),
+    email: yup.string().email().required('Email is required'),
+    password: yup.string().required('Password is required'),
```
jamessspanggg, line 57: should we also include the password strength regex here so that the user knows the password strength criteria while typing in their password? currently, the criteria only shows after the user issues the post request:
create

jamessspanggg (commented):
Tested the sign in and register functionality and both worked fine. 💯

However, here are some initial observations:
-  for the sign in page, when the user just finished typing the password, he/she cannot directly click on the login button. Only when the user clicks on some other component, the login button is clickable:
login
- should we auto log in the user after successful creation of an account?

src/containers/CreateAccount/index.tsx
buckyroberts: There are some rules that are difficult to program into the frontend such as the `UserAttributeSimilarityValidator` and `CommonPasswordValidator` https://docs.djangoproject.com/en/3.1/topics/auth/passwords/ - I wasn't sure how to handle those and didn't want our FE and Be to get out of sync so I just let the BE throw back the errors.

buckyroberts:
@jamessspanggg I am having trouble reproducing the "for the sign in page, when the user just finished typing the password, he/she cannot directly click on the login button." issue - what browser are you using?

src/containers/SignIn/index.tsx
buckyroberts: updated

jamessspanggg:
> @jamessspanggg I am having trouble reproducing the "for the sign in page, when the user just finished typing the password, he/she cannot directly click on the login button." issue - what browser are you using?

Chrome Version 86.0.4240.111

buckyroberts:
@jamessspanggg I see what's happening. Chrome is autofilling the email field so to Formik that field appear not `touched`. Looking for a solution now...

jamessspanggg:
> @jamessspanggg I see what's happening. Chrome is autofilling the email field so to Formik that field appear not `touched`. Looking for a solution now...

Just investigated the issue, it seems like it is due to the `validation` not being run at every `onChange`, even though we have the `validateOnChange` property as true.

Observe that the screen recording below where while I am typing the input, the `errors` on the right and `isValid` does not update on change:
formik-validation

PR #1169, opened 2021-01-02
Add validateField as callback during onChange

### LINKED ISSUE #1168: Auth forms: unable to click on submit button after filling all inputs correctly
**Bug Description**
The submit button is still disabled when a user tries to click on the submit button right after filling all the inputs correctly. 

**Steps to Reproduce**
1. Go to Sign in page
2. Fill in any email
3. Fill in any password and directly click on the submit button
4. View that the submit button is not clickable

**Expected behavior**
After filling in all input details, the submit button should be clickable.

**Actual behavior**
Submit button is still disabled after filling in correct input details.

**Screenshots/Recordings**
Note that right after the user fills in all the details, the submit button is disabled. The submit button is only enabled after the user clicks on something else.
login_button

**OS and Browser**
 - OS: Mac, Chrome.

**Additional context**
This is probably due to the form input elements do not validate based on change. From the recording below, the errors object seems to not update on every change.
formik-validation

**Account Number**

jamessspanggg:
Fixes #1168 

From debugging the behaviour of formik on our website, it seems like the onChange events do not update the `errors` object as well as the `isValid` property. This causes the error messages to not update based on each input from the user, as well as the issue #1168.

This PR calls the `validateField` function to the onChange handler, which updates the `errors` object and `isValid` property every time the `onChange` event is fired.

Note: I'm not too familiar with Formik and unsure if this is the best and most elegant solution.

### DIFF
--- a/src/components/FormComponents/FormInput/index.tsx
+++ b/src/components/FormComponents/FormInput/index.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable react/jsx-props-no-spreading */
 
-import React, {FC} from 'react';
+import React, {ChangeEvent, FC} from 'react';
 import {Field} from 'formik';
 import clsx from 'clsx';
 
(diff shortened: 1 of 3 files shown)


src/components/FormComponents/FormInput/index.tsx
```diff
@@ -19,13 +19,20 @@ const FormInput: FC<ComponentProps> = ({
   ...baseInputProps
 }) => {
   const {className, name} = baseInputProps;
-  const {errors, touched} = useFormContext();
+  const {errors, handleChange, touched, validateField} = useFormContext();
   const error = !!errors[name] && !!touched[name];
 
+  const handleChangeAndValidate = (e: ChangeEvent) => {
+    handleChange(e);
+    setTimeout(() => {
```
jamessspanggg, line 27: https://github.com/formium/formik/issues/2059#issuecomment-612733378

thesanjeevsharma (approved):

PR #1180, opened 2021-01-05
Made mobile drop down icon clickable

### LINKED ISSUE #1179: Mobile drop down icon not clickable
**Bug Description**
Clicking TopNavMobileMenu's drop down icons do not toggle show/hide. Clicking only the title will toggle show/hide.

**Steps to Reproduce**
1. Go to website home page with mobile device
2. Click on top right hamburger icon
3. Click on the drop down icon 
4. View that nothing happens

**Expected behavior**
Clicking on drop down icon should show/hide items

**Actual behavior**
Nothing happened 

**Screenshots/Recordings**
https://user-images.githubusercontent.com/32864116/103612713-f8d30b00-4f5f-11eb-8c03-2ad520814900.MP4

**OS and Browser**
 - OS: iOS safari

**Additional context**

**Account Number**

AkinAguda:
Fixes #1179 

I changed the `div` with the `TopNavMobileMenu__icon-holder` className to a `button` with the `TopNavMobileMenu__icon-button` className. This is because an `onClick` was being assigned was to be assigned to that element which makes it now actt as a button.

I removed any default button style associated with that button so ensure it looks exactly like it did when it was still a div.

### DIFF
--- a/src/data/teams.json
+++ b/src/data/teams.json
@@ -736,7 +736,20 @@
           "profileImage": "https://avatars1.githubusercontent.com/u/55309498?s=460&u=5735170037f5398148b87115ee9c9e7df574b7b1&v=4",
           "slackUsername": "Siddhesh Kulthe"
         }
-      }
+      },
+      {
+          "title": "Blog Project Manager",
+          "isLead": true,
+          "payPerDay": 2800,
+          "createdDate": "2020-07-08T02:14:59.307535Z",
+          "contributor": {
+            "contributorId": "63",
+            "displayName": "Ijeoma Okoh",
+            "githubUsername": "Aiijay",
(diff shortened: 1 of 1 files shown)


src/containers/TopNav/TopNavMobileMenu/index.tsx
```diff
@@ -41,14 +41,14 @@ const TopNavMobileMenu = () => {
           {title}
         </button>
         {smallDevice && (
-          <div className="TopNavMobileMenu__icon-holder">
```
jamessspanggg, line 44: Instead of making both the title and icon clickable individually, how about just make the outer div clickable?

AkinAguda: Okay then. I'll do that instead

PR #1284, opened 2021-02-13
[#1283] Add blog link to top nav and footer

### LINKED ISSUE #1283: Add link to blog on website
- [ ] Add links to the blog (https://thenewboston.blog/) on both the top and bottom nav bars on the website. One under the 
the More menu and one in the footer. 

https://www.figma.com/file/UbsJyxNB7ruowSqjpx43pP/Blogs?node-id=0%3A1

jamessspanggg:
Fixes #1283 

Things to note:
- Added a `isExternal` boolean to the links to take into account external urls
- Conditionally render between React Router `Link` and HTML anchor link `A` based on `isExternal`, as I believe `Link` cannot handle external urls.

Acc no: c54c04774efaae20746fd3f29cdd619fea512e119bc1082b863572b2e8844104

### DIFF
--- a/src/components/A/A.scss
+++ b/src/components/A/A.scss
@@ -1,7 +1,5 @@
 .A {
   .Icon {
     display: inline-flex;
-    margin-left: 1px;
-    margin-right: 3px;
   }
 }
--- a/src/components/A/A.scss
+++ b/src/components/A/A.scss
@@ -1,9 +1,7 @@
 .A {
-  .Icon {
+  // Needed to add extra specificity to override `Icon`'s styling
+  .A__Icon--new-window {
     display: inline-flex;
-  }
-
-  &__Icon--new-window {
     margin-left: 1px;
     margin-right: 3px;
   }
--- a/src/components/A/A.scss
+++ b/src/components/A/A.scss
@@ -2,4 +2,9 @@
   .Icon {
     display: inline-flex;
   }
+
+  &__Icon--new-window {
+    margin-left: 1px;
+    margin-right: 3px;
+  }
 }
--- a/src/components/A/index.tsx
+++ b/src/components/A/index.tsx
@@ -4,12 +4,13 @@ import clsx from 'clsx';
 interface ComponentProps {
   className?: string;
   href: string;
+  onClick?: React.MouseEventHandler<HTMLAnchorElement>;
   target?: string;
 }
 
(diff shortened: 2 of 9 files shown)


angle943:
I am currently talking with the team about this Issue, I will keep you updated @jamessspanggg

angle943 (changes requested):
@jamessspanggg After discussing with Bucky and Kristy, we decided to remove the Criteria `Should open a new window`. So we **don't** want this link to open a new window, but rather just change the page within the same window (the reason for this is we generally don't want to open a new window for a button, especially when all buttons around it doesn't open a new window).

So let's do the following:

1) first pull in the latest changes, as I have recently updated the `A` component. (you'll probably get merge conflicts, be sure to discard your changes on the `A` component due to my next step:
2) remove the `onClick` for the `A` tag, since we no longer need that (since we are just changing the page to the blog page, we don't care we don't close the popup)
3) when using the (updated) `A` tag, be sure to pass in the prop `newWindow={false}`, because we don't want it to open a new window.

Let me know if that works. If new issues arises, or you have some concerns, please leave comments on this PR. thanks!

src/components/A/A.scss
```diff
@@ -1,7 +1,5 @@
 .A {
   .Icon {
     display: inline-flex;
-    margin-left: 1px;
```
jamessspanggg, line 4: hey @angle943, I have implemented based on the new criterias. However, since there were margins added on your previous commit, it was interfering with the margins in the `TopNavPopoverItem` components, basically your newly added `.A .Icon` styles has a higher specificity. 

Screen Shot 2021-02-14 at 10 33 24 PM

I removed it for now, as it was not impacting the existing UI much? But would be open if there's a better way to maintain both UIs.

jamessspanggg: Update: 
I added the margins back to the `.A .Icon` styles, however, I made it more specific by adding the `--new-window` modifier.

```
A {
   // property must be placed in `.Icon` instead of `&__Icon--new-window` to 
   // prevent original `.Icon` display flex style to override this
  .Icon {
    display: inline-flex;
  }

  &__Icon--new-window {
    margin-left: 1px;
    margin-right: 3px;
  }
}
```

angle943: ah that `.Icon{...}` was an oversight in my part. Shouldn't have made the styling that way

PR #1285, opened 2021-02-13
added four contributors 

dhakal0kushal:


jamessspanggg:
@thesanjeevsharma is it okay for this to be in a single PR since there's so many outstanding "New Contributor" issues? Might be redundant to create them separately

dhakal0kushal:
I think it will be okay for the pull request to be single since we can use issues to track the payments, not the PRs.

PR #1383, opened 2021-02-21
1373 - add tests for SlideUp component

### LINKED ISSUE #1373: UNIT TESTS - components/SlideUp
## What

Create unit tests using [Jest](https://jestjs.io/en/) and [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/).

Here are a couple examples of how unit tests should be created:

1. [Unit tests for our A component](https://github.com/thenewboston-developers/Website/blob/development/src/components/A/A.test.tsx)
2. [Unit tests for Icon component](https://github.com/thenewboston-developers/ui-js/blob/master/src/components/Icon/Icon.test.tsx)
3. [Unit tests for Button component](https://github.com/thenewboston-developers/ui-js/blob/master/src/components/Inputs/Button/Button.test.tsx)

## General Acceptance Criteria for writing Unit tests

1. Should create a `[nameOfComponent].test.tsx` within the same directory of the component.
2. Should have tests that tests every `prop` of the component.
3. If there are any conditional logic within the component, there should be unit tests that accounts for each conditional branches.
4. `npm run test` should be passing.

## Specific Acceptance Criteria

1. make sure you tests for each prop

PatNeedham:
closes #1373 

image

image

The three lines at the beginning of the test file setting up the div#slide-up-root element are copied from [the docs](https://testing-library.com/docs/example-react-modal/) (changing the div's ID)

### DIFF
--- a/src/components/SlideUp/SlideUp.test.tsx
+++ b/src/components/SlideUp/SlideUp.test.tsx
@@ -17,12 +17,13 @@ describe('SlideUp component', () => {
   test('calls close prop when component is clicked', () => {
     const close = jest.fn();
     render(<SlideUp close={close}>Hello World</SlideUp>);
-    fireEvent.click(screen.getByTestId('SlideUp'));
-    expect(close).toHaveBeenCalledTimes(1);
+    fireEvent.click(screen.getByTestId('SlideUp__overlay'));
+    expect(close).toHaveBeenCalled();
   });
 
   test('renders with className passed in', () => {
     render(<SlideUp close={() => {}} className="test" />);
-    expect(screen.getByTestId('SlideUpContent')).toHaveClass('test__content');
+    expect(screen.getByTestId('SlideUp')).toHaveClass('test');
+    expect(screen.getByTestId('SlideUp__content')).toHaveClass('test__content');
   });
 });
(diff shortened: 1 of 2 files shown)


src/components/SlideUp/SlideUp.test.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+describe('SlideUp component', () => {
+  test('renders with text passed in', () => {
+    render(<SlideUp close={() => {}}>Hello World</SlideUp>);
+    expect(screen.getByText('Hello World')).toBeTruthy();
+  });
+
+  test('calls close prop when component is clicked', () => {
+    const close = jest.fn();
+    const {getByRole} = render(<SlideUp close={close}>Hello World</SlideUp>);
```
angle943, line 19: lets not do this destructuring from render, and grabbing by the Role. You can pass an `data-testid` if you want to grab the button.

angle943 (changes requested):
Can you add tests for when passing along className? Make sure the `bemify(...)` part gets tested too

src/components/SlideUp/SlideUp.test.tsx
PatNeedham: Thank you for the suggestion. I'm changing it to that in the next git push arriving shortly.

angle943:
I've updated the `data-testid`'s to have parity with the classNames

PR #1385, opened 2021-02-21
test: added tests for Avatar component

thesanjeevsharma:
closes: #1334 

I had some problem spreading props because of eslint. Should we allow that?

Also, this is me writing tests for the first time. 😬 So, let me know if there's any scope for improvement.


angle943:
looks good to me! Good job @thesanjeevsharma on the thoroughness!

I've updated to allow for prop-spreading on all our test files. Also I updated the className of the fallback component ( i didn't like how the className for that one was set)

PR #1294, opened 2021-02-15
pr: run tests

shashanksinha89:

PR #1388, opened 2021-02-21
Add github actions test workflow

jamessspanggg:
Fixes part of https://github.com/thenewboston-developers/DevOps/issues/6

Seems like https://github.com/thenewboston-developers/Website/pull/1294 is abandoned and there were errors. It would be good to set this up now as there will be more PRs coming in to create tests. Adding test to the CI pipeline will save effort from reviewers to ensure tests have no failure.

Things to note:
- I'm only running the tests for node v12 and v14, I believe it's unnecessary to run node v10 and adding that will just take up more processing time for github actions (note that there are [usage limits](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration) to github actions). 
  - We can perhaps only run on one node version, which we can specify in the `README` file. This is so that CI runs faster and less prone to failures due to a lot of PRs trying to run github workflows at the same time.
- Currently the workflow will only execute when:
  -  it's a PR to the development branch, as well as pushes to the PR
  - any direct pushes to the team's repo

References:
- https://github.com/yarnpkg/yarn/issues/4890 (why i added the network timeout in `yarn install --network-timeout 1000000`)

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,0 +1,43 @@
+name: Test
+on:
+  push:
+  pull_request:
+    branches:
+      - development
+jobs:
+  build:
+    name: Running test on Node ${{ matrix.node }} and ${{ matrix.os }}
+
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        node: ['12.x', '14.x']
+        os: [ubuntu-latest, windows-latest, macOS-latest]
+
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v2
+
+      - name: Use Node ${{ matrix.node }}
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node }}
+
+      - name: Cache node modules
+        uses: actions/cache@v2
+        env:
+          cache-name: cache-node-modules
+        with:
+          # npm cache files are stored in `~/.npm` on Linux/macOS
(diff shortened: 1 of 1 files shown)


.github/workflows/testing.yml
```diff
@@ -0,0 +1,43 @@
+name: Test
+on:
+  push:
+  pull_request:
+    branches:
+      - development
```
thesanjeevsharma, line 6: You haven't specified any branches under push. Will it consider `development` for push and pull_request if I write it like this?

jamessspanggg: When branches aren't specified, then all branches will be considered, including development.

Reason that I allowed push on all branches is because it allows subsequent pushes on pull requests to run the workflow. (The pull_request only runs when u make the pr, not when subsequent pushes to the pr)

thesanjeevsharma: Are you sure about this? Our Lint action is listening on master and development for pushes and PRs. Albeit, I have noticed that if you push to any existing PR it will re-run Lint action it.

jamessspanggg: My bad you are right, for some reason when I was testing it previously, it doesn't run on push to PRs. I have changed accordingly. 😅

thesanjeevsharma (approved):

angle943:
nice job on being proactive about this @jamessspanggg !

PR #1400, opened 2021-02-22
Test for ContributorTasks

tomijaga:
solves: #1339
account number: 8de30226230c35bbc1ce4a63c62a7b9c86bf0ce21fc7bc1a984b7884a9f88782

### DIFF
--- a/src/components/ContributorTasks/ContributorTasks.test.tsx
+++ b/src/components/ContributorTasks/ContributorTasks.test.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable react/jsx-props-no-spreading */
 import React from 'react';
 import {render, screen} from '@testing-library/react';
 import '@testing-library/jest-dom/extend-expect';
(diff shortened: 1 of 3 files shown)


jamessspanggg (commented):
One comment from me

src/components/ContributorTasks/index.tsx
```diff
@@ -6,34 +6,38 @@ import {A} from 'components';
...
   tasks: Task[];
 }
 
-const ContributorTasks: FC<ComponentProps> = ({className, tasks}) => {
+const ContributorTasks: FC<ContributorTasksProps> = ({className, tasks}) => {
   const renderRows = () => {
     return tasks.map(({amount_paid, completed_date, issue_id, repository, title}: Task, index) => (
-      <tr key={index}>
-        <td className="ContributorTasks__task-title" title={title}>
+      <tr key={index} data-testid="ContributorTasks--row">
```
angle943, line 17: lets change this name to `ContributorTasks__row` 

( __ denotes an element, while -- denotes a modifier)

src/components/ContributorTasks/ContributorTasks.test.tsx
```diff
@@ -0,0 +1,124 @@
+/* eslint-disable react/jsx-props-no-spreading */
```
angle943, line 1: can you pull in the latest changes from `development`, and then remove this `eslint-disable`?

src/components/ContributorTasks/ContributorTasks.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+import React from 'react';
+import {render, screen} from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import format from 'date-fns/format';
+
+import ContributorTasks, {ContributorTasksProps} from './index';
+
+describe('ContributorTasks', () => {
+  const props: ContributorTasksProps = {
```
angle943, line 10: you can move this `props` outside of the `describe`. Put it in between the last `import` (on line 7) and before the `describe`

src/components/ContributorTasks/ContributorTasks.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+    render(<ContributorTasks {...props} />);
+
+    expect(screen.getByTestId('ContributorTasks')).toBeTruthy();
+  });
+
+  it('renders with default className', () => {
+    render(<ContributorTasks {...props} />);
+    const taskComp = screen.getByTestId('ContributorTasks');
+
+    expect(taskComp).toHaveClass('ContributorTasks');
```
angle943, line 44: can you add more `expect`s here? test for the other DOM elements that you gave `data-testid` to (if they have a default className)

src/components/ContributorTasks/ContributorTasks.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+    expect(validTitle).toBeTruthy();
+  });
+
+  it('renders rows with valid github issue link', () => {
+    render(<ContributorTasks {...props} />);
+    const tasks = screen.getAllByTestId('A');
+
+    const validClass = tasks.every((task) => {
+      return task.className.includes('ContributorTasks__issue-link');
+    });
```
angle943, line 76: I think we can remove this test here checking for the correct className

src/components/ContributorTasks/ContributorTasks.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+  it('renders rows with valid amount', () => {
+    render(<ContributorTasks {...props} />);
+    const tasks = screen.getAllByTestId('ContributorTasks__amount');
+
+    const validAmount = tasks.every((task, i) => {
+      const amount = props.tasks[i].amount_paid;
+      const formattedAmount = parseInt(amount, 10).toLocaleString();
+      return task.textContent === `+ ${formattedAmount}`;
+    });
```
angle943, line 120: hmm can we actually test this explicitly writing our what the string should be (instead of using logic to create the string)?

The reason is this... one of the benefits of unit tests is that it is almost self-documenting our code. But if we add too much JS logic within the test, we lose the self-documentation aspects of it.

Another reason is that it shows us exactly what the results should look like. A future developer might come in and make edits to this component, sees that this test is failing, and then just copies and pastes the logic to the test. This might defeat the purpose of the test later down the road

src/components/ContributorTasks/ContributorTasks.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+  });
+
+  it('renders rows with date compeleted', () => {
+    render(<ContributorTasks {...props} />);
+    const tasks = screen.getAllByTestId('ContributorTasks__date-completed');
+
+    const validDate = tasks.every((task, i) => {
+      const date = format(props.tasks[i].completed_date, 'L/d/yy');
+      return task.textContent === date;
+    });
```
angle943, line 107: let's test the dates with explicit strings (look at my long comment below)

src/components/ContributorTasks/ContributorTasks.test.tsx
```diff
@@ -0,0 +1,124 @@
...
+    expect(validHref).toBeTruthy();
+  });
+
+  it('renders rows with valid github repository repository', () => {
+    render(<ContributorTasks {...props} />);
+    const tasks = screen.getAllByTestId('ContributorTasks__repository');
+
+    const validRepo = tasks.every((task, i) => {
+      return task.textContent === props.tasks[i].repository;
+    });
```
angle943, line 95: let's test this with explicit strings (look at my long comment below)

angle943 (changes requested):

angle943 (approved):

PR #1401, opened 2021-02-22
Test for DocList

tomijaga:
solves: #1348
account number: 8de30226230c35bbc1ce4a63c62a7b9c86bf0ce21fc7bc1a984b7884a9f88782

### DIFF
--- a/src/components/DocWrapper/DocList/DocList.test.tsx
+++ b/src/components/DocWrapper/DocList/DocList.test.tsx
@@ -1,0 +1,62 @@
+/* eslint-disable react/jsx-props-no-spreading */
+import React from 'react';
+import {render, screen} from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+
+import DocList, {DocListProps} from './index';
+
+describe('DocList', () => {
+  let props: DocListProps;
+
+  beforeEach(() => {
+    props = {
+      className: 'test-class',
+      variant: 'ul',
+    };
+  });
+
+  it('renders without crashing', () => {
+    render(<DocList {...props}>Hello World</DocList>);
+  });
+
+  it('renders with proper default className', () => {
+    render(<DocList {...props}>Hello World</DocList>);
+    const list = screen.getByTestId(`DocList--${props.variant}`);
+
+    expect(list).toHaveClass('DocList');
+  });
+
+  it('renders unorderd list when specified in variant prop', () => {
+    render(<DocList {...props}>Hello World</DocList>);
+    const list = screen.getByTestId(`DocList--ul`);
+
+    expect(list).toBeTruthy();
+    expect(list).toHaveClass(`DocList--ul`);
+  });
+
+  it('renders unorderd list with passed in className', () => {
+    render(<DocList {...props}>Hello World</DocList>);
+    const list = screen.getByTestId(`DocList--ul`);
+
+    expect(list).toHaveClass('test-class');
+  });
+
+  it('renders orderd list when specified in variant prop', () => {
+    render(<DocList variant="ol">Hello World</DocList>);
+    const list = screen.getByTestId(`DocList--ol`);
+
+    expect(list).toBeTruthy();
+    expect(list).toHaveClass(`DocList--ol`);
+  });
+
+  it('renders orderd list with passed in className', () => {
+    render(
+      <DocList className={props.className} variant="ol">
+        Hello World
+      </DocList>,
+    );
+    const list = screen.getByTestId(`DocList--ol`);
+
+    expect(list).toHaveClass('test-class');
+  });
+});
--- a/src/components/DocWrapper/DocList/DocList.test.tsx
+++ b/src/components/DocWrapper/DocList/DocList.test.tsx
@@ -39,6 +39,7 @@ describe('DocList', () => {
     const list = screen.getByTestId(testIdUl);
 
     expect(list).toHaveClass('test-class');
+    expect(list).toHaveClass('test-class--ul');
   });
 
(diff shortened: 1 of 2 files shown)


src/components/DocWrapper/DocList/DocList.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+describe('DocList', () => {
+  let props: DocListProps;
+
+  beforeEach(() => {
+    props = {
+      className: 'test-class',
+      variant: 'ul',
+    };
+  });
```
thesanjeevsharma, line 16: What's the benefit of this? 🤔 If you need the same props every time, why use `beforeEach`?

src/components/DocWrapper/DocList/DocList.test.tsx
```diff
@@ -0,0 +1,62 @@
+/* eslint-disable react/jsx-props-no-spreading */
```
thesanjeevsharma, line 1: You can remove this. Take a pull from development. Test files have been allowed props spreading.

thesanjeevsharma (changes requested):

src/components/DocWrapper/DocList/DocList.test.tsx
tomijaga: I fixed it and just initialized the props variable with the values

src/components/DocWrapper/DocList/DocList.test.tsx
```diff
@@ -0,0 +1,57 @@
+import React from 'react';
+import {render, screen} from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+
+import DocList, {DocListProps} from './index';
+
+describe('DocList', () => {
+    const props: DocListProps = {
+    className: 'test-class',
```
jamessspanggg, line 9: The indentation here should be two spaces after `const`

src/components/DocWrapper/DocList/DocList.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    variant: 'ul',
+    };
+
+  it('renders without crashing', () => {
+    render(<DocList {...props}>Hello World</DocList>);
+  });
+
+  it('renders with proper default className', () => {
+    render(<DocList {...props}>Hello World</DocList>);
+    const list = screen.getByTestId(`DocList--${props.variant}`);
```
jamessspanggg, line 19: I think you can create a `const testIdUl = 'DocList--ul` and `const testIdOl = 'DocList--ol` and reuse them all over

src/components/DocWrapper/DocList/DocList.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+    expect(list).toHaveClass('DocList');
+  });
+
+  it('renders unorderd list when specified in variant prop', () => {
```
jamessspanggg, line 24: *unordered

src/components/DocWrapper/DocList/DocList.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    render(<DocList {...props}>Hello World</DocList>);
+    const list = screen.getByTestId(`DocList--ul`);
+
+    expect(list).toBeTruthy();
+    expect(list).toHaveClass(`DocList--ul`);
+  });
+
+  it('renders unorderd list with passed in className', () => {
```
jamessspanggg, line 32: here too

src/components/DocWrapper/DocList/DocList.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    render(<DocList {...props}>Hello World</DocList>);
+    const list = screen.getByTestId(`DocList--ul`);
+
+    expect(list).toHaveClass('test-class');
+  });
+
+  it('renders orderd list when specified in variant prop', () => {
```
jamessspanggg, line 39: here too

src/components/DocWrapper/DocList/DocList.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+    render(<DocList variant="ol">Hello World</DocList>);
+    const list = screen.getByTestId(`DocList--ol`);
+
+    expect(list).toBeTruthy();
+    expect(list).toHaveClass(`DocList--ol`);
+  });
+
+  it('renders orderd list with passed in className', () => {
```
jamessspanggg, line 47: here

jamessspanggg (commented):
@angle943 @thesanjeevsharma 

I have a suggestion. Should we create a data-testId variable in the original component file. i.e. `export const dataTestId = ...` and the test file can just import it. This will ensure re-usability and decrease the mistake of passing the `testId` to `getByTestId` wrongly

angle943:
> @angle943 @thesanjeevsharma
> 
> I have a suggestion. Should we create a data-testId variable in the original component file. i.e. `export const dataTestId = ...` and the test file can just import it. This will ensure re-usability and decrease the mistake of passing the `testId` to `getByTestId` wrongly

@jamessspanggg hmm the only problem with that is, what if we have more than one data-testid in a component? Also, `data-testid`s should have same name as `classNames`, so why don't we export the className too?

But there might be some refactors we can do in the future regarding the fact that classNames and data-testid have parity... hmm let's think about this for a bit

src/components/DocWrapper/DocList/DocList.test.tsx
angle943: also please move the `const props` out of the describe please

angle943 (approved):

PR #1402, opened 2021-02-23
ADD: Test for SocialMediaIcon;

denistsoi:
Resolves: https://github.com/thenewboston-developers/Website/issues/1374

I've purposefully left out the test case where website is null. 
(as typescript will lint against this since it's not a valid enum in `SocialMedia`).

I've also included get by classname as well - Thx


angle943:
@denistsoi let's keep the tests a unit test, meaning they should be testing just one thing. Also we should be testing for all props, including the `className` prop

angle943 (approved):

PR #1409, opened 2021-02-24
test for GoToTop

### LINKED ISSUE #1356: UNIT TESTS - components/GoToTop
## What

Create unit tests using [Jest](https://jestjs.io/en/) and [React Testing Library](https://testing-library.com/docs/react-testing-library/intro/).

Here are a couple examples of how unit tests should be created:

1. [Unit tests for our A component](https://github.com/thenewboston-developers/Website/blob/development/src/components/A/A.test.tsx)
2. [Unit tests for Icon component](https://github.com/thenewboston-developers/ui-js/blob/master/src/components/Icon/Icon.test.tsx)
3. [Unit tests for Button component](https://github.com/thenewboston-developers/ui-js/blob/master/src/components/Inputs/Button/Button.test.tsx)

## General Acceptance Criteria for writing Unit tests

1. Should create a `[nameOfComponent].test.tsx` within the same directory of the component.
2. Should have tests that tests every `prop` of the component.
3. If there are any conditional logic within the component, there should be unit tests that accounts for each conditional branches.
4. `npm run test` should be passing.

## Specific Acceptance Criteria

angle943:
closes #1356

### DIFF
--- a/src/components/GoToTop/GoToTop.test.tsx
+++ b/src/components/GoToTop/GoToTop.test.tsx
@@ -1,0 +1,24 @@
+import * as React from 'react';
+import {render, screen} from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+
+import GoToTop from '.';
+
+describe('GoToTop', () => {
+  afterEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('renders without crashing', () => {
+    render(<GoToTop />);
+  });
+
+  it('displays icon when visible', () => {
+    jest.spyOn(React, 'useState').mockImplementation(() => [true, jest.fn()]);
+
+    render(<GoToTop />);
+    const button = screen.queryByTestId('GoToTop');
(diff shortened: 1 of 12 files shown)


src/components/GoToTop/GoToTop.test.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
```
angle943, line 1: Pretty annoying, but I had to `import * as React from 'react'` and not `import React from 'react'` because of the `jest.spyOn` on line 17.

angle943: more detail on this. This is due to how we are importing `useState` on the actual component file.

Over there, we are importing it like `import React, {useState, etc etc} from 'react'`.

Due to this, the `spyOn` will not recognize the `useState`

thesanjeevsharma (approved):

PR #1410, opened 2021-02-24
chore: update test

thesanjeevsharma:
Updated test to use explicit strings instead of JS hacks.


angle943:
@thesanjeevsharma eh actually I'm not so sure about this one. Maybe it was fine how you had it haha

thesanjeevsharma:
Then we can close this. @angle943 

Throws Computer

PR #1414, opened 2021-02-25
top nav mobile update

angle943:
Two things I wanted to fix for mobile view regarding our top nav:

1. The texts are all grey, making it seem like it's disabled
2. Our topNav component has a foggy-glass look, but the menu dropdown is pure white. This makes it look off.

Left is how our app currently looks like, and right is my proposed change:

Screen Shot 2021-02-24 at 6 22 44 PM

Basically, the foggy-glass look will disapear when the dropdown is open.

I added a bunch of reviewers to this, to hopefully ensure that my change doesn't yield something unintended

### DIFF
--- a/src/components/Layout/Layout.scss
+++ b/src/components/Layout/Layout.scss
@@ -10,8 +10,7 @@ $footer-height: 348px;
   }
 
   &__content,
-  &__Footer,
-  &__TopNav {
+  &__Footer {
     @media (min-width: 1366px) {
       margin-left: auto;
       margin-right: auto;
@@ -29,8 +28,6 @@ $footer-height: 348px;
   }
 
   &__top-nav-wrapper {
-    backdrop-filter: saturate(180%) blur(20px);
-    background: rgba(255, 255, 255, 0.8);
     position: sticky;
     top: 0;
     z-index: 1;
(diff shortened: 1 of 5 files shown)


amarjeet987 (approved):
Yeah seems like the issues is fixed. Now that we're on the nav view, should we do something for the hamburger menu as well and get it done with. It turns grey on click for mobile and only reverts back when clicked on somewhere else in the screen, seems a bit of a weird experience, not really sure what the grey highlight represents here as a user. @angle943 what do you suggest?

angle943:
@amarjeet987 good call. We should blur once thats clicked. Let me add something for that

amarjeet987:
> @amarjeet987 good call. We should blur once thats clicked. Let me add something for that

Yeah we can make it grey when it is "active" and white by default I guess.

angle943:
@amarjeet987 updated to account for that button

jamessspanggg:
I remember we had a fix where the slide down animation should be sliding down under (in terms of z-index) the navbar. Currently it is on top of the navbar, i.e. text are showing on the navbar.

navbar

thesanjeevsharma (approved):
LGTM, just the animation issue that @jamessspanggg mentioned ^

angle943:
i undid my change to use `portals` to revert the z-index issues

PR #1419, opened 2021-02-27
ADD: Test for Loader;

denistsoi:
Resolves: https://github.com/thenewboston-developers/Website/issues/1361

### DIFF
--- a/src/components/Loader/Loader.test.tsx
+++ b/src/components/Loader/Loader.test.tsx
@@ -1,0 +1,44 @@
+import React from 'react';
+import {render, screen} from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+
+import Loader from './index';
+
+describe('Loader component', () => {
+  test('renders with testId Loader', () => {
+    render(<Loader />);
+    const loader = screen.getByTestId('Loader');
+    expect(loader).toBeTruthy();
+  });
+
+  test('renders with default className', () => {
+    render(<Loader />);
+    const loader = screen.getByTestId('Loader');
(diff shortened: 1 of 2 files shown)


src/components/Loader/Loader.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  test('renders with LoadingIcon', () => {
+    render(<Loader />);
+    const loader = screen.getByTestId('Loader');
+    expect(loader).toBeTruthy();
+
+    const icon = loader.querySelector('.mdi-icon');
+    expect(icon).toBeTruthy();
+    expect(icon?.getAttribute('width')).toEqual('15.35');
```
jamessspanggg, line 41: Can switch to this so that you don't need to optional chain it:

`expect(icon.style).toHaveProperty('height', '15.35px');`

jamessspanggg (commented):
One nit from me:

src/components/Loader/Loader.test.tsx
denistsoi: hmmm, 
getting an error if i try the suggestion with `querySelector` - 
not sure if i can query `.mdi-icon` without affecting the `LoadingIcon`

```
  expect(received).toHaveProperty(path, value)

    Expected path: "height"

    Expected value: "15.35"
    Received value: ""
```

denistsoi: I guess i can pass a data-testId - one second

jamessspanggg (approved):

PR #1423, opened 2021-02-27
Remove contributor

bojan-ilijevski:


bojan-ilijevski:
It's resolved in #1425

PR #1443, opened 2021-03-01
test: added tests for Layout component

thesanjeevsharma:
closes: #1360


src/components/Layout/Layout.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const portal = document.createElement('div');
+    portal.setAttribute('id', 'popover-root');
+
+    document.body.appendChild(portal);
+
+    window.scrollTo = scrollToFn;
+    window.HTMLElement.prototype.scrollIntoView = scrollIntoViewFn;
+  });
+
+  afterAll(cleanup);
```
angle943, line 39: do we need this? I tried deleting it, and the tests still passed.

it looks like its supposed to be run automatically: https://testing-library.com/docs/react-testing-library/api#cleanup

angle943 (approved):
everything looks good, but i think we can get rid of `cleanup` (and the associated import)

PR #1455, opened 2021-03-03
ADD: MenuGroup Tests;

denistsoi:
Resolves: https://github.com/thenewboston-developers/Website/issues/1364


src/components/MenuGroup/MenuGroup.test.tsx
```diff
@@ -0,0 +1,96 @@
+import React from 'react';
+import {BrowserRouter as Router} from 'react-router-dom';
+import {render, screen, fireEvent} from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+
+import MenuGroup from './index';
```
thesanjeevsharma, line 6: import MenuGroup from '.';

src/components/MenuGroup/MenuGroup.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+describe('MenuGroup component', () => {
+  test('renders without crashing', async () => {
```
thesanjeevsharma, line 9: Do you require `async` here? Try removing it.

src/components/MenuGroup/MenuGroup.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+    render(
+      <Router>
+        <MenuGroup title="Test" urlBase="Test" />
```
thesanjeevsharma, line 12: No passing props like this. Create an object before `describe` and use that. Just like I pointed out in the last PR.

Check this commit:
[Tabs tests](https://github.com/thenewboston-developers/Website/commit/5a9442bb197d32a455d35dac5b0ad61beb3a0f78)

src/components/MenuGroup/MenuGroup.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+      </Router>,
+    );
+
+    expect(screen.getByTestId('MenuGroup')).toBeTruthy();
+  });
+
+  test('renders with customTestId "Custom" without crashing', async () => {
```
thesanjeevsharma, line 19: Not really an issue, but still, use `it` instead of `test`.

src/components/MenuGroup/index.tsx
```diff
@@ -10,21 +10,28 @@ import './MenuGroup.scss';
 interface ComponentProps extends RouteComponentProps {
   title: string;
   urlBase: string;
+  dataTestId?: string;
```
thesanjeevsharma, line 13: Don't do this.

src/components/MenuGroup/index.tsx
```diff
@@ -10,21 +10,28 @@ import './MenuGroup.scss';
...
 }
 
-const MenuGroup: FC<ComponentProps> = ({children, location, title, urlBase}) => {
+const MenuGroup: FC<ComponentProps> = ({children, dataTestId = 'MenuGroup', location, title, urlBase}) => {
   const [expanded, toggleExpanded] = useState(getFirstPathParam(location.pathname) === urlBase);
 
   return (
-    <div className="MenuGroup">
+    <div className="MenuGroup" data-testid={dataTestId}>
```
thesanjeevsharma, line 20: Just write data-testid="MenuGroup" here.

src/components/MenuGroup/MenuGroup.test.tsx
thesanjeevsharma: Also, remove this test

src/components/MenuGroup/MenuGroup.test.tsx
```diff
@@ -0,0 +1,96 @@
...
+    expect(screen.getByText('Test')).toBeTruthy();
+  });
+
+  test('renders with child components', async () => {
+    render(
+      <Router>
+        <MenuGroup title="Test" urlBase="Test">
+          <div>Child 1</div>
+          <div>Child 2</div>
+        </MenuGroup>
```
thesanjeevsharma, line 44: See if you can add a prop `children` and use that instead. You will have to update index.tsx too.

thesanjeevsharma (changes requested):
Whatever comments I have left apply to all the tests.

src/components/MenuGroup/index.tsx
denistsoi: I’ll need it for this PR

https://github.com/thenewboston-developers/Website/pull/1461

src/components/MenuGroup/index.tsx
denistsoi: Same as above comment 
https://github.com/thenewboston-developers/Website/pull/1461

src/components/MenuGroup/MenuGroup.test.tsx
denistsoi: https://www.thenewboston.com/style-guide/react#prop-spreading

src/components/MenuGroup/MenuGroup.test.tsx
denistsoi: I’ll fix that, must I’ve been a bad copy paste sorry, 🙏

src/components/MenuGroup/MenuGroup.test.tsx
denistsoi: Do we really need to type check children as React.node? 

Children is a default prop in React.FC

src/components/MenuGroup/MenuGroup.test.tsx
thesanjeevsharma: No problem

src/components/MenuGroup/MenuGroup.test.tsx
thesanjeevsharma: It's allowed while writing tests. 😄 

Thanks for bringing this up. I'll update the docs.

src/components/MenuGroup/MenuGroup.test.tsx
thesanjeevsharma: I know, but this is the pattern we've been following. If multiple tests use children. You could just update it in one place.
```
const baseProps : MenuGroupProps = {
  children: <h1>Hello</h1>
}
```

src/components/MenuGroup/index.tsx
thesanjeevsharma: I think you can test that, without a test id too. It'll not be pretty tho. You'll have to use methods like `getByRole()`, `queryByText()` etc.

@angle943 thoughts?

angle943: so this is certainly not the recommended way in the `react-testing-library` methodology, but I've been breaking a lot of their rules too to be honest(i'm too used to using `enzyme`).

Because I've been doing this in other areas, I think for now we can allow it. I may revisit our tests as a whole in the future and refine our tests to fit more with `react-testing-libraries`'s philosophy.

But at least until we have an end-to-end testing set up (via Cypress), I think we can be relaxed with what RTL recommends

src/components/MenuGroup/MenuGroup.test.tsx
denistsoi: Ok no worries, just wanted to raise it before I edit it after dinner tonight.

src/components/MenuGroup/index.tsx
denistsoi: I found it quite difficult to query (without testid) unless I use something querySelectorAll and iterate through those items for navigationData... 

Agreed it wouldn’t be pretty either but that’s an alternative if we want to be strict

src/components/MenuGroup/MenuGroup.test.tsx
denistsoi: Ok no probs. I’ll do it when I get home tonight

denistsoi:
I used `role={role}` and added role="Group" to still get MenuGroup

thesanjeevsharma (approved):

PR #1465, opened 2021-03-05
[#1464] Provide better UX for Forms

### LINKED ISSUE #1464: Better UX for forms
- Show loading icon when users are submitting the forms.
- Show disabled button if the inputs are invalid.

jamessspanggg:
Fixes #1464

- Loading icon will be shown when user is submitting the form
- Submit button will be disabled when inputs are invalid
- Styled the disabled button to look grey-ish so that the button will look disabled-ish

Sign in form:

https://user-images.githubusercontent.com/32864116/110145370-4fe84580-7e14-11eb-9a9f-7820d41b9f0b.mp4

Create account form:

https://user-images.githubusercontent.com/32864116/110145991-00eee000-7e15-11eb-9a4b-98b533fa307c.mp4

### DIFF
--- a/src/containers/SignIn/index.tsx
+++ b/src/containers/SignIn/index.tsx
@@ -20,10 +20,13 @@ const SignIn: FC = () => {
   const dispatch = useDispatch<AppDispatch>();
   const history = useHistory();
   const [errorMessage, setErrorMessage] = useState<string>('');
+  const [submitting, setSubmitting] = useState<boolean>(false);
 
   const handleSubmit = async ({email, password}: FormValues): Promise<void> => {
     try {
+      setSubmitting(true);
       const {data} = await dispatch(login({email, password}));
+      setSubmitting(false);
       history.push(`/users/${data.user.pk}`);
     } catch (error) {
       setErrorMessage(formatAPIError(error));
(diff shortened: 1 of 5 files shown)


angle943 (changes requested):
When I try logging in, and I typed in a wrong password, the error came up properly, but the button was stuck in loading:

Screen Shot 2021-03-06 at 7 59 09 PM

src/containers/SignIn/index.tsx
```diff
@@ -20,10 +20,13 @@ const SignIn: FC = () => {
...
+  const [submitting, setSubmitting] = useState<boolean>(false);
 
   const handleSubmit = async ({email, password}: FormValues): Promise<void> => {
     try {
+      setSubmitting(true);
       const {data} = await dispatch(login({email, password}));
+      setSubmitting(false);
       history.push(`/users/${data.user.pk}`);
     } catch (error) {
       setErrorMessage(formatAPIError(error));
```
angle943, line 32: you can remove the `setSubmitting(false)` from line 29, and instead, below the `catch` block, put a `finally { setSubmitting(false); }`

jamessspanggg: Don't think putting in the `finally` block is the best as the `setSubmitting(false)` will be after `history.push(...)`. That will lead to a unable to setState after component is unmounted error.

I added another `setSubmitting(false)` to the `catch` block instead

angle943: yeah good point

angle943 (approved):

PR #1499, opened 2021-03-08
Fix console error `validateDOMNesting`

### LINKED ISSUE #1479: Console error: validateDOMNesting
**Bug Description**
When going through some pages in the documentation, the following console errors appear: `validateDOMNesting(...): <div> cannot appear as a descendant of <p>`.

**Steps to Reproduce**
1. Run development server and visit localhost
2. Go to documentation
3. Click on Style Guide -> React
4. View console erorrs

**Expected behavior**
No console errors should occur.

**Actual behavior**
Console errors occurred.

**Screenshots/Recordings**
Screen Shot 2021-03-07 at 11 20 40 AM

**OS and Browser**
 - OS: MacOS Big Sur
 - Browser: Chrome

**Additional context**

**Account Number**

elvisiraguha:
Fix #1479 

Account: b4c1c1ba27b0535a642fd2d00f64da64447826979db36051b6354d8336458722

### DIFF
--- a/src/components/Icon/index.tsx
+++ b/src/components/Icon/index.tsx
@@ -110,14 +110,14 @@ interface ComponentProps {
   dataTestId?: string;
   disabled?: boolean;
   icon: IconType;
-  onClick?(e?: React.MouseEvent<HTMLDivElement, MouseEvent>): void;
-  onKeyDown?(e?: React.KeyboardEvent<HTMLDivElement>): void;
+  onClick?(e?: React.MouseEvent<HTMLSpanElement, MouseEvent>): void;
+  onKeyDown?(e?: React.KeyboardEvent<HTMLSpanElement>): void;
   size?: number;
   totalSize?: number | 'unset';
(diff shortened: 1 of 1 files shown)


jamessspanggg (commented):
Hmm, this indeed solves the problem, but it would be good to use `<p>` to follow semantic HTML. I would suggest changing to use href `<a>` tags instead of using our `<A>` component.

elvisiraguha:
> Hmm, this indeed solves the problem, but it would be good to use `<p>` to follow semantic HTML. I would suggest changing to use href `<a>` tags instead of using our `<A>` component.

Using `<a>` instead of `<A>` would omit the new window icon as shown below

using `<A>`
Screen Shot 2021-03-11 at 14 18 55

using `<a>`
Screen Shot 2021-03-11 at 14 18 35


But I think this can be solved in the `<Icon>` implementation, instead of using a div, we can use a span.

Issue #1489, opened 2021-03-07
Add a required `Display Name: field` in the sign up form


angle943:
Below `Confirm Password` field, make a required `display name` field. `name=display_name`

PR #1507, opened 2021-03-09
Add Display Name field in the sign up

manishram:
Fix: https://github.com/thenewboston-developers/Website/issues/1489

Account Number: `0f92e975101eb96cea80dce430d74efd3172fa229377818d467ded85dcf2ab6c`

### DIFF
--- a/src/containers/CreateAccount/index.tsx
+++ b/src/containers/CreateAccount/index.tsx
@@ -21,11 +21,15 @@ const CreateAccount: FC = () => {
   const [errorMessage, setErrorMessage] = useState<string>('');
   const [submitting, setSubmitting] = useState<boolean>(false);
 
-  const handleSubmit = async ({email, password}: FormValues): Promise<void> => {
+  const handleSubmit = async ({display_name, email, password}: FormValues): Promise<void> => {
     try {
       setErrorMessage('');
       setSubmitting(true);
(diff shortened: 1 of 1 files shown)


amarjeet987 (commented):
@angle943 Can the user use numbers / special characters in the display name?
If not, @manishram you can put up some html pattern kinda thingy like "[a-zA-Z]*"  for the input to prevent the user from using numbers and special characters in the display name.

manishram:
I thought that too but then realised maybe people would like to use display names like **Chri$to4** instead of **Christopher**. 😂

 Let's see what Justin say.

angle943:
Good points, for now, lets just do these two:

1) no whitespaces on the left & right (we can probably just `trim` the field before sending it off to the BE)
2) max length of `30` (instead of using `yup` for this, we can probably just use `max_length` on the `Input`. Or if our `Input` is not configured to accept that, you can use `yup` validation for that

angle943 (changes requested):
look at comments above

jamessspanggg:
Added the trimming of display name before sending to BE, also added the 30 character limit with yup. Our current `Input` does not handle it.

cc @angle943 @manishram

manishram:
Thanks, @jamessspanggg I got busy with Weekly Reports.

angle943 (approved):

manishram:
Hey, @jamessspanggg can you give me your acc # so that I can send you 250coins for helping me out in this PR.

Issue #1528, opened 2021-03-12
UX Designer & Researcher


kiezela:
Add the following job opening to the website.

***

#### Job Title
UX Designer & Researcher

#### Job Description
Responsible for researching, categorizing, and documenting user behavior by employing design methodologies, facilitating internal & external discussions, and visualizing concepts. Ensure design concepts are fully realized by working closely with software developer teams and collaborating with other designers to arrive at the best design solution.

#### Responsibilities

- Conduct competitor analysis, customer analysis and user research.

- Attend to product structure and strategy, as well as content development.

- Develop user flows, prototypes and wireframes.

- Coordination with other designers and developers.

- Conduct user testing and analysis.

- Continuously iterate on solutions to make the product increasingly effective and delightful for users.

#### Skills and Qualifications

- Collaboration and communication

- Critical thinking and problem-solving skills

- Familiarity with user research and analysis technique

- Wireframing and prototyping

- Understanding of information architecture

- User and usability testing

- Successful completion of UX/UI course work or training portfolio

- Proven experience at a professional level

#### Reports To
- First Name: [Kristy]
- Slack Name: [Kristy]

#### Pay
- 2,800 coins/day

#### Application channels
- Slack - Message [Kristy]
- LinkedIn
- Twitter
- Reddit

Tatenda:
I'm working on this

Tatenda:
@jamessspanggg

angle943:
paid

PR #1559, opened 2021-03-15
UX Designer & Reasearch #1528

Tatenda:
Issue #1528  Account Number: 3449835feb55f29b3df19e1d0cee5aacfafded42beb1b2146fe1fc31fa2bf3cd

### DIFF
--- a/src/data/openings.json
+++ b/src/data/openings.json
@@ -1,421 +1,427 @@
-[
-  {
-    "openingId": "senior-python-django-developer",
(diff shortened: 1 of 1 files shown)


amarjeet987 (changes requested):
Please always make sure all checks are successful while making a PR, otherwise the changes won't be merged. Can you look into why the checks are failing? Do let me know if you need any help with that.

PR #1570, opened 2021-03-16
fix/package-lock

thesanjeevsharma:
node version: 12.18.3
npm version: 6.14.6


thesanjeevsharma:
@angle943 after updating node to `v14.16.0` and npm to `v7.6.3`, I still have a few changes. But I think these are okay?

PR #1571, opened 2021-03-16
added transition instead simple pop on openings

rupamkairi:
## Added transition effect on the Openings page 

Added transition on OpeningsOpening class.

https://user-images.githubusercontent.com/33358148/111327064-1248af80-8693-11eb-98e4-06cd38c4ca39.mp4

Hope this is not against your styleguides.


amarjeet987:
Is it related to some reported issue?

amarjeet987 (commented):
Is it related to some reported issue?

rupamkairi:
No, I don't think it was mentioned in any issue. I did this as a visual improvement.

angle943:
i like it, except we have a lot of cards that should all have consistent UX experience, so closing for now

PR #1574, opened 2021-03-18
hide all openings

angle943:


src/utils/data.ts
```diff
@@ -33,8 +33,8 @@ export const getTasks = (): TaskDict => {
   return results;
 };
 
-export const getTeamMemberByGithubUsername = (github_username: string): TeamMember | undefined => {
-  let member: TeamMember | undefined;
+export const getTeamMemberByGithubUsername = (github_username: string): TeamMember | null => {
+  let member: TeamMember | null = null;
```
jamessspanggg, line 37: hmmm what does `null = null` do here?

angle943: its saying that `member`, which can be of type `TeamMember` or `null`, will be initialized as `null`

angle943: i basically added this because when we `return` the `member` variable at the end of this function, my IDE was complaining that it might not be initialized

PR #1583, opened 2021-03-22
Adding Whole new blog with path /blog

IvanDanilov66:
@angle943 Here's the new blog.


jamessspanggg:
@IvanDanilov66 seems like there are lint errors with ur PR. Also, we are not using `styled-components` to style our components, we are using `BEM` CSS to do so (i.e. separate SCSS files, you can check the existing code: https://github.com/thenewboston-developers/Website/tree/development/src/components/ContributorTasks).

IvanDanilov66:
No worries I am working on it we will fix the lint error and change the styled-components.

PR #1644, opened 2021-03-30
[#1643] Disable automatic scroll restoration

### LINKED ISSUE #1643: Scroll restoration causes content jump when reloading page
https://user-images.githubusercontent.com/32864116/112998400-983f1d00-91a0-11eb-8993-862812413cde.mp4

jamessspanggg:
Fixes #1643 

Browser compatibility: https://developer.mozilla.org/en-US/docs/Web/API/History#browser_compatibility (it does not support IE but i guess its fine since we dont support IE anyways)

After:

https://user-images.githubusercontent.com/32864116/112998995-22878100-91a1-11eb-9219-085b6a98cf93.mp4

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,72 +1,74 @@
 <!DOCTYPE html>
 <html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
+    <meta name="theme-color" content="#000000" />
+    <meta name="description" content="We are building the future of open source digital currency. Join the team!" />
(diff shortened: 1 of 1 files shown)


thesanjeevsharma (approved):
This seems to be an issue(feature for long pages, maybe?) with Chrome. Firefox doesn't restore it by default.

This works but let's get@angle943 's view.

angle943 (approved):

angle943:
@jamessspanggg what was the change? Its hard to tell from the diff

public/index.html
```diff
@@ -1,72 +1,74 @@
...
-         gtag('config', 'UA-56989641-1', { send_page_view: false });
+        gtag('js', new Date());
+        gtag('config', 'UA-56989641-1', {send_page_view: false});
       }
-   </script>
+    </script>
 
-</head>
+    <!-- Disable automatic scroll restoration -->
+    <script>
```
jamessspanggg, line 43: @angle943 here

PR #1653, opened 2021-03-31
Update hello-world.ts

gunars-b:


angle943:
hello, we are no longer taking hello world PRs

PR #1663, opened 2021-04-03
[#1590] Migrate teams page to use website API

jamessspanggg:
Fixes #1590 

**Progress**
- [x] Hook up with necessary BE apis for Teams page and call in FE
- [x] Hook up with necessary BE apis for Profile page and call in FE
- [x] Remove teams related data
- [x] Create dummy data and populate to `api-dev`
- [x] Confirm with BE on the (1) `slack` field on `Team` and (2) if `responsibilities` field should be an array of strings

**Before Merging**
- [x] Check that `api` has all the necessary Team, CoreTeam, CoreTeamMembers data

**Fixed some UI issues along the way as well:**
<details>
<summary>
Fixed all team members flashing before team filters are even set
</summary>

https://user-images.githubusercontent.com/32864116/114305890-97957780-9b0c-11eb-93c7-11d5bd3d1fbe.mp4

</details>

PR #1704, opened 2021-04-13
add mrsky and update some other things

angle943:


angle943:
1) cleaned up the logic. We had it where a user acn be involved with multiple teams, and thus have multiple titles. We got rid of this rule, so I cleaned it up to be singular values for both.

2) added the ability to filter the team member page via the position, using queryParams. I did this because from the `Projects` page, I want to be able to link to just the project reviewers

src/containers/Teams/index.tsx
```diff
@@ -87,26 +91,18 @@ const Teams: FC = () => {
...
-            teamLeads.push({...member, isLead: true});
-          } else {
-            otherMembers.push({...member, isLead: false});
-          }
-        }
-      });
-      return teamLeads.concat(otherMembers);
-    };
+    const getFilteredMembers = () =>
+      teamMembers.filter((teamMember) => teamMember.team.toLowerCase() === teamFilter.toLowerCase());
```
jamessspanggg, line 95: is the order of `lead` and normal team members not matter anymore?

src/containers/Teams/index.tsx
```diff
@@ -87,26 +91,18 @@ const Teams: FC = () => {
...
+
     const members = teamFilter === TeamName.all ? teamMembers : getFilteredMembers();
     const sortedMembers = sortTeamMembers(members);
-    setFilteredMembers(sortedMembers);
-  }, [teamFilter]);
+
+    if (queryTitle) {
+      setFilteredMembers(sortedMembers.filter((member) => member.title.toLowerCase() === queryTitle.toLowerCase()));
```
jamessspanggg, line 101: what is this change for? how can we set the `queryTitle` within our website? These changes are causing quite a few merge conflicts to https://github.com/thenewboston-developers/Website/pull/1663 but not sure to follow this implementation

jamessspanggg: what's the use case for this? it seems like only if we add `title` query param manually in the url, then we will filter the teams based on member

jamessspanggg: ok saw ur explanation above, ignore this

src/containers/Teams/index.tsx
angle943: what do you mean?

thesanjeevsharma: @jamessspanggg On line 98, there's `sortTeamMembers` function. That takes care of it.

PR #1711, opened 2021-04-15
Add configs to toggle showing of auth buttons

### LINKED ISSUE #1710: Hide Create Account/Sign In buttons until beta is out
@angle943 Hey Justin, can we do something super quick and hide the Create Account/Sign In buttons until registered username is out? I dont want to lead people to clicking Create Account when we're not ready for it yet.

jamessspanggg:
Fixes #1710 

Adding config variables would be a good option to toggle on and off certain functionalities

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -1,0 +1,2 @@
+export const isCreateAccountAllowed = false;
+export const isSignInAllowed = false;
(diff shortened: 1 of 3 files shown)


src/config/index.ts
```diff
@@ -0,0 +1,2 @@
+export const isAllowCreateAccount = false;
+export const isAllowSignIn = false;
```
thesanjeevsharma, line 2: @jamessspanggg 
I think a better name would be `isCreateAccountAllowed` or `isCreateAccountVisible`.

Same for sign-in.

jamessspanggg: yep changed

thesanjeevsharma (approved):

PR #1733, opened 2021-04-21
[#1702] Font UI Update

### LINKED ISSUE #1702: AppWide UI Update - Update Headers
here is our new style guide: https://www.figma.com/file/6AGSvP6DJIIvy5Ayyp1a6U/Design-System?node-id=12%3A4

as you can see from that figma link, we will have standards when it comes to headers (h1,h2,h3,h4) and what they are calling `Display`.

Find an elegant way to tackle this issue app-wide. Make sure you test each page that will be effected to ensure that it will not be broken

jamessspanggg:
Fixes #1702 

Based on latest specification: https://www.figma.com/file/6AGSvP6DJIIvy5Ayyp1a6U/Design-System?node-id=2650%3A49

Note: there might be a slight change in UI with this PR, might need more testing for this. Might need to request designer's review as well


src/styles/_font.scss
```diff
@@ -35,12 +42,40 @@
...
-  --font-family-mono: JetBrainsMono, monospace;
-  --font-size-large: 24px;
-  --font-size-medium: 20px;
-  --font-size-regular: 16px;
-  --font-size-small-regular: 14px;
-  --font-size-small: 12px;
+  --font-family-jetbrains: JetBrainsMono, monospace;
+
+  // font weights
+  --bold: 700;
```
jamessspanggg, line 50: changed all the font variables based on the specifications

src/styles/_font-mixins.scss
```diff
@@ -0,0 +1,76 @@
+%display1 {
```
jamessspanggg, line 1: created a font mixins file so that the fonts are easily extended

jamessspanggg:
@angle943 I have added the new font design system to the website. For the existing one that I'm not sure of, I just used the same values (but using the css font variables that I have created instead of magic values)

Note that for some reason, previously the `font-weight: 700` fonts are thinner than the ones at `font-weight: 600 or 500`. (due to So you might see some fonts being thicker than previous as now there have the correct `font-weight: 700` thickness.

We can merge this in first, then slowly correct the fonts if the designers are not satisfied. Let me know if this is ok.

New font system (tested it out locally and cross checked with the design specifications): 
Screen Shot 2021-04-22 at 8 58 22 PM

PR #1852, opened 2021-06-12
Remove x-scroll in downloads page

### LINKED ISSUE #1839: x-scroll issue
There is a x-scroll on the window when it should be responsive instead. Tested on the download page.

Screen Shot 2021-06-09 at 9 47 30 PM

avinashupadhya99:
Closes #1839 

Account number - `94cf53921e9c5ddf48dc5589b7cdf111f7acdc93dc3073b94880416d4676b9ae`

### DIFF
--- a/src/containers/Download/Download.scss
+++ b/src/containers/Download/Download.scss
@@ -1,6 +1,7 @@
 @import 'styles/font-mixins';
 
 $top-nav-height: 60px;
+$scroll-bar-width: 20px;
 
 .Download {
   display: flex;
@@ -18,6 +19,7 @@ $top-nav-height: 60px;
     background-color: var(--color-gray-050);
     display: flex;
     justify-content: center;
+    width: calc(100vw - #{$scroll-bar-width});
   }
 
   &__main-content {
--- a/src/containers/Download/Download.scss
+++ b/src/containers/Download/Download.scss
@@ -5,6 +5,7 @@ $top-nav-height: 60px;
 .Download {
   display: flex;
   flex-direction: column;
+  overflow-x: hidden;
   position: relative;
 
   @media (max-width: 480px) {
--- a/src/containers/Download/Download.scss
+++ b/src/containers/Download/Download.scss
@@ -5,7 +5,6 @@ $top-nav-height: 60px;
 .Download {
   display: flex;
   flex-direction: column;
-  overflow-x: hidden;
   position: relative;
 
   @media (max-width: 480px) {
@@ -19,7 +18,6 @@ $top-nav-height: 60px;
     background-color: var(--color-gray-050);
     display: flex;
     justify-content: center;
-    width: 100vw;
   }
 
   &__main-content {


jamessspanggg:
@angle943 can u review this if you can replicate this issue? for some reason i can't replicate this issue in my local

jamessspanggg:
@avinashupadhya99 instead of setting `overflow-x: hidden` to solve the issue, can you figure out which component is overflowing?

mtrierweiler:
The problem with just removing `width:100vw` Is that the background color of the downloads will no longer span the screen leaving the default background color in the margins. I believe this was the purpose of the Downloads__main-background. That way it has some contrast to the release notes section.

avinashupadhya99:
@mtrierweiler That's my bad, sorry. Did not notice the background color did not span. I'll look into another way of fixing it. Thanks

avinashupadhya99:
The issue is with the `width: 100vw` on `Downloads__main-background`. According to this [SO answer](https://stackoverflow.com/a/30504010), it is not possible to have 100vw and not have a horizontal scrollbar when there is a vertical scrollbar. I can set the overflow-x to hidden for  Downloads__main-background after setting the width back to 100vw. Please let me know if this works

mtrierweiler:
I don't believe that will work. Because setting .Dowload__main-background to overflow-x:hidden will hide any of the children that overflow but I believe it won't get rid of the scroll bar. Have you tried it?

avinashupadhya99:
I have tested this approach of excluding the scroll bar width from the `Dowload__main-background`'s width in Chrome, Edge and Firefox. I'm not sure about Safari

jamessspanggg:
dont think its the best solution as we can't guarantee that the y scrollbar is 20px, can you try this instead?

avinashupadhya99:
Setting the max-width to 100% does not increase the size of the `Download__main-background` div to 100vw. I feel like the 20px scroll bar fix works in most browsers, I have checked on Chrome, Edge and Firefox.

Here's a screenshot with the max-width set to 100% and the width set to 100vw - 
image

jamessspanggg:
> Setting the max-width to 100% does not increase the size of the `Download__main-background` div to 100vw. I feel like the 20px scroll bar fix works in most browsers, I have checked on Chrome, Edge and Firefox.

Yeah but you are only testing with a Windows machine i assume? different OS-es have different scrollbar widths.

If this is the case, I would prefer the initial solution of setting `overflow-x: false`, since we know that the problem is due to the  scrollbar width and `100vw`

avinashupadhya99:
> Yeah but you are only testing with a Windows machine i assume? different OS-es have different scrollbar widths.

I did not know that.

> If this is the case, I would prefer the initial solution of setting `overflow-x: hidden`, since we know that the problem is due to the scrollbar width and `100vw`

Unfortunately, setting the overflow-x to hidden on the Download class or any parent class/div will not stretch the width of `Download__main-background` to 100vw. I had not noticed this earlier.

angle943:
closing this one. We are going to be having a lot of pages in the future where there will be parts with different bg color. I don't want to put in a temporary hacky fix to solve this issue. The fix I have in mind is quite elaborate, and may require me to go into every component and update something. I will work on this issue

PR #1860, opened 2021-06-13
Homepage social icons lag issue Fixed

roovy54:
Account Number: `daa957ba0a1b63d5d724bc6e11553dbd6981d7d5da4a4a3daf4fd79c67058c8f`

### DIFF
--- a/src/containers/Home/HomeHero/HomeHero.scss
+++ b/src/containers/Home/HomeHero/HomeHero.scss
@@ -127,6 +127,7 @@ $title-width: 592px;
     height: 64px;
     justify-content: center;
     width: 64px;
+    transition: background-color 0s, color 0s;
 
     &:not(:last-child) {
       margin-right: 40px;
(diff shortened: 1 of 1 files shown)


src/containers/Home/HomeHero/HomeHero.scss
```diff
@@ -127,6 +127,7 @@ $title-width: 592px;
     height: 64px;
     justify-content: center;
     width: 64px;
+    transition: background-color 0s, color 0s;
```
jamessspanggg, line 130: hi can you explain why this addition will solve the issue? I don't see any other places that would affect the transition here

roovy54: While inspecting the website, The default transition was 0.3s and so if we change it to 0s the lag gets fixed.

Home _ thenewboston - Google Chrome 14-06-2021 21_33_13 (2)_LI

jamessspanggg (approved):

roovy54:
Can I order the properties in alphabetical order.


build_prob

PR #1879, opened 2021-06-18
Add blurred image placeholder for HomeHero image

jamessspanggg:
Created component `ProgressiveImage`. Need to supply both a tiny version of the original image and the original image to the component. 

Would suggest to convert the original image to `webp` format via https://squoosh.app/

Screen recording below shows before and after:

https://user-images.githubusercontent.com/32864116/122581655-dfdf9180-d089-11eb-9abf-dea86d44a218.mp4

https://user-images.githubusercontent.com/32864116/122628614-afc7db00-d0e9-11eb-901b-20217d3174d4.mp4

Inspiration from Medium posts. 

References:
- https://jmperezperez.com/webp-placeholder-images/
- https://jmperezperez.com/medium-image-progressive-loading-placeholder/


src/containers/Home/HomeHero/index.tsx
```diff
@@ -62,9 +64,14 @@ const HomeHero: FC = () => {
             <div className="HomeHero__social-media-links">{renderSocialMediaLinks()}</div>
           </div>
         </div>
-        <div className="HomeHero__right">
-          <img alt="hero" className="HomeHero__image" src={HeroV2Svg} />
-        </div>
+        {width > 414 && (
```
thesanjeevsharma, line 67: @jamessspanggg Do we need this condition?

jamessspanggg: Previously, it sets the `display: none` when the screen width is less than 414 px. The reason why I added this condition is so that we will not issue the network request to obtain the images if so

thesanjeevsharma (commented):
LGTM ✔️ 

Also, @jamessspanggg how about renaming the component to `ProgressiveImage`?

jamessspanggg:
> LGTM ✔️
> 
> Also, @jamessspanggg how about renaming the component to `ProgressiveImage`?

sure that's a better name, let me change it

thesanjeevsharma (approved):

src/components/ProgressiveImage/index.tsx
```diff
@@ -0,0 +1,62 @@
...
+  height,
+  placeholderImageClassName,
+  placeholderSrc,
+  realImageClassName,
+  realSrc,
+  width,
+}: Props) => {
+  const [isLoaded, setIsLoaded] = useState(false);
+  return (
+    <div className={clsx('ProgressiveImage', containerClassName)} style={{height, width}}>
```
angle943, line 29: Ideally it would be cool if we can use this `ProgressiveImage` component just like how we use ``. Is there a reason why this component needs this outer `div`? Without it it can behave much more like a native `img`

jamessspanggg: This component needs an outer `div` as there are two overlapping images, one for the placeholder and one for the real image. Both of them need to be `position: absolute` to overlap each other, which means that there needs to be a relatively positioned outer div.

angle943: ah makes sense

PR #1894, opened 2021-06-25
100vw fix

### LINKED ISSUE #1890: 100vw fix
I will work on this

angle943:
Our new designs will have sections where the background will have 100vw, but the way that our code was structured prevented things from being 100vw, since we dictated a `max-width` in the `Layout` component (which encompasses all components).

This forced us to come up with hacky solutions such as explicitely putting styling to override this.

Because I see this a common occurrence as design gets updated, I decided that it is no longer appropriate for our app to have a `max-width` automatically enforced to all components. Rather, I have created a `Container` component that naturally works like a Container (ex: https://material-ui.com/components/container/)

I have gone through every single page on our website to make sure that it incorporates it, and I have also applied changes where it applies.

closes #1890

### DIFF
--- a/src/components/Container/index.tsx
+++ b/src/components/Container/index.tsx
@@ -1,0 +1,42 @@
+/* eslint-disable react/jsx-props-no-spreading */
+
+import React, {FC, useMemo} from 'react';
+import clsx from 'clsx';
+
+import './Container.scss';
+
+type ContainerElement = 'div' | 'header' | 'main';
+
+interface ContainerProps {
+  className?: string;
+  element?: ContainerElement;
+  dataTestId?: string;
+  maxWidth?: number;
+}
+
+const Container: FC<ContainerProps> = ({children, className, dataTestId, element = 'div', maxWidth = 1366}) => {
+  const props = useMemo(
+    () => ({
+      className: clsx('Container', className),
+      'data-testid': dataTestId,
+      style: {
+        maxWidth,
+      },
+    }),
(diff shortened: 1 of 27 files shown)


src/containers/Projects/ProjectRulesAndGuide/index.tsx
```diff
@@ -8,6 +8,7 @@ import Rules from './Rules';
 import ProposalSubmissionProcess from './ProposalSubmissionProcess';
 import MilestonesAndPayouts from './MilestonesAndPayouts';
 import './ProjectRules.scss';
+import {Container} from '../../../components';
```
jamessspanggg, line 11: can use absolute imports instead of relative

src/components/Container/index.tsx
```diff
@@ -0,0 +1,42 @@
+/* eslint-disable react/jsx-props-no-spreading */
+
+import React, {FC, useMemo} from 'react';
+import clsx from 'clsx';
+
+import './Container.scss';
+
+type ContainerElement = 'div' | 'header' | 'main';
```
thesanjeevsharma, line 8: should we have `section` too? 🤔

thesanjeevsharma (approved):
LGTM

src/components/Container/index.tsx
angle943: yeah we can have all the elements. But i just added the ones that were already used & in need of this component. Feel free to add more to this list in the future whenever you feel the need

PR #1897, opened 2021-06-26
Fix case insensitivity for faq searches

### LINKED ISSUE #1884: Search queries should be case insensitive
**Bug Description**
When doing a search, an exact match should not be required, and a search should work regardless of the search query case.

**Steps to Reproduce**
Steps to reproduce the behaviour:
1. Go to FAQ page (https://thenewboston.com/faq/all)
2. Type "coins" (for example) and hit search. You will see some results. Now change the case of any letter (Coins or coinS or CoinS).
3. No results will be shown

**Expected behaviour**
The search should display results regardless of the case of the query

**Actual behaviour**
Results are shown only when a query is entirely in lower case

**OS and Browser**
 - OS: Windows
 - Browser: All

**Additional context**

**Account Number**
18c07b4227cb0993fa7a1db834834e42461a0d9f61b2b157dfbfee625bbf3de6

avinashupadhya99:
Closes #1884 

Account number - `94cf53921e9c5ddf48dc5589b7cdf111f7acdc93dc3073b94880416d4676b9ae`

### DIFF
--- a/src/containers/Faq/index.tsx
+++ b/src/containers/Faq/index.tsx
@@ -51,7 +51,7 @@ const Faq: FC = () => {
       .map((faqTopic) => {
         return {
           // filter questions that match with search query
-          content: faqTopic.content.filter((qna) => qna.question.toString().toLowerCase().includes(searchQuery)),
+          content: faqTopic.content.filter((qna) => qna.question.toString().toLowerCase().includes(sanitizedSearch)),
           topic: faqTopic.topic,
         };
       })


src/containers/Faq/index.tsx
```diff
@@ -51,7 +51,7 @@ const Faq: FC = () => {
       .map((faqTopic) => {
         return {
           // filter questions that match with search query
-          content: faqTopic.content.filter((qna) => qna.question.toString().toLowerCase().includes(searchQuery)),
+          content: faqTopic.content.filter((qna) => qna.question.toString().toLowerCase().includes(sanitizedSearch)),
```
angle943, line 54: nice!

PR #1923, opened 2021-07-04
[#1922] Remove roadmap from ProjectDetails

### LINKED ISSUE #1922: Remove roadmap in project details
Screen Shot 2021-07-04 at 4 44 22 PM

jamessspanggg:
Fixes #1922


src/containers/Projects/ProjectDetails/constants.ts
```diff
@@ -33,12 +33,6 @@ export const projectDetailsTopic: ProjectTopicMap = {
     position: 1,
```
thesanjeevsharma, line 33: Hey, `position` is not being used now I think. Can we also get rid of these in this PR?

jamessspanggg: It is being used by the `ProjectDetailsSideMenu` component which uses the ordered version of the project topics.

thesanjeevsharma: 🤔 that could have been achieved w/o position too, if we disable eslint rule on that file?

thesanjeevsharma:
@jamessspanggg there's an icon for the roadmap. Should we remove that too?

src/containers/Projects/ProjectDetails/constants.ts
angle943: @thesanjeevsharma technically its unreliable to sort objects, although in practice it'll always be ordered the way we want it (after we disable es-lint). But just out of principle, i never like depending on the order of an object alone to dictate it's order

angle943 (approved):

Issue #1928, opened 2021-07-07
Add blog link to navigation (top and footer)


manishram:
Add blog link: https://blog.thenewboston.com to the navigation (top and footer) of the main site.

**Here is the Figma design:**
https://www.figma.com/file/UbsJyxNB7ruowSqjpx43pP/Blogs?node-id=308%3A667

![blog](https://user-images.githubusercontent.com/22790904/124728865-8ef6e680-df2d-11eb-8888-d03cedcc25f6.png)

PR #1929, opened 2021-07-07
feat: add blog link in footer and top nav

thesanjeevsharma:
closes: #1928

The blog opens on the same tab right now. I think it would be better if we could open a new tab for it. I tried using `newWindow`, but it doesn't blend well with the rest of the nav links. It also adds extra space because of the icon(image below).
If a new tab is not required then this can be merged, else let me know.

image

### DIFF
--- a/src/components/Footer/index.tsx
+++ b/src/components/Footer/index.tsx
@@ -46,6 +46,11 @@ const navLists = [
         title: 'Community Guidelines',
         url: '/guidelines',
       },
+      {
+        isExternal: true,
+        title: 'Blog',
+        url: 'https://blog.thenewboston.com',
+      },
     ],
   },
   {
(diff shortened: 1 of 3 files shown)


jamessspanggg:
I think we can just use react router link to open in a new tab without the icon?

manishram:
Hey @jamessspanggg,
Our blog site resides in a subdomain so there is **no problem opening link in the same tab**. Google started considering subdomains as internal linking since 2011 update.

manishram (approved):

src/components/Footer/index.tsx
```diff
@@ -46,6 +46,11 @@ const navLists = [
         title: 'Community Guidelines',
         url: '/guidelines',
       },
+      {
+        isExternal: true,
```
jamessspanggg, line 50: can remove `isExternal: true` since it can be internal link?

jamessspanggg (approved):

Issue #1931, opened 2021-07-09
Add Developer Portal's Living Whitepaper's Landing Page onto the website


ma2b0043:
Hi, you need to add Developer Portal's Living Whitepaper's Landing Page onto the website. The things you may need to complete this task are as follows:
- Content:
https://docs.google.com/document/d/1_6OCmqL0OK5UiWJROe-EbRYAQa9s-iuIIQOdv_Vd5Bs/edit
- UI & Handoff
https://www.figma.com/file/x6gQN0dfrzCq1iDfbLQTXZ/Living-Whitepaper?node-id=1%3A3
- Other materials
https://drive.google.com/drive/folders/1WZYNsT2bJ16_zbteY34gd2HHfIPeyZUr?usp=sharing
White Paper Diagrams-20210705T032731Z-001.zip

Link to design issue: https://github.com/thenewboston-developers/Design/issues/223

All of the design guidelines and materials are provided in the links listed above^, and if you need assistance, kindly let us know asap, so that we can help you.

Breakdown of tasks (to be confirmed):
- [x] Add basic routes for pages (James)
- [x] Create dev portal layout, contains top links, breadcrumbs container, side bar etc (James)
- [x] Responsive main structure of the living whitepaper landing page (James)
- [x] Principal Entities on the network documentation (Sanjeev)
- [x] Principal Events and Processes on the network documentation (Sanjeev)
- [x] Architecture - deep dive documentation (James)
- [x] Breadcrumb component + sidebar component (James)

jamessspanggg:
Hi, can provide permissions for the content as well as the UI & handoff to the FE team?

manishram:
@jamessspanggg please DM me your Figma email on Discord. I'll send you the invite.

jamessspanggg:
hi @angle943 @thesanjeevsharma 

Since Justin is occupied with the beta wallet, Sanjeev should we work on this together? This page seems to be quite huge.

We can break the task down, make small changes and merge to development regularly, since the page is not able to be accessed by the public.

thesanjeevsharma:
Sure @jamessspanggg, let me have a thorough look at the Figma design then I'll get back to you on discord. 👍

PR #1944, opened 2021-07-14
feat/white paper principal entities

thesanjeevsharma:
Part of #1931


src/containers/LivingWhitepaper/LivingWhitepaper.scss
```diff
@@ -110,7 +118,17 @@
...
     background-color: var(--color-cyan-900);
-    padding: 48px 72px;
+    padding: 72px 48px;
+    padding-right: 72px;
+
+    @media (max-width: 768px) {
+      padding-right: 48px;
+    }
+
+    @media (max-width: 768px) {
```
jamessspanggg, line 128: hmm seems like this media query is the same as the one above? should it be `414px` or smth?

src/containers/LivingWhitepaper/containers/PrincipalEntities/index.tsx
```diff
@@ -1,10 +1,595 @@
...
+            Registered users choice for governor
+          </p>
+        </section>
+        <Divider className="PrincipalEntities__divider" />
+        <div className="PrincipalEntities__bottom-bar">
+          <div
+            className="PrincipalEntities__bottom-bar-content"
+            onClick={() => history.push('/developer/whitepaper/principal-events')}
+            role="button"
+            tabIndex={0}
```
jamessspanggg, line 583: can we make this a component under `LivingWhitepaper/components`? so we can reuse in other pages. Can pass in the URL and text to be displayed as props

jamessspanggg (commented):
Sick work, the page looks amazing.

src/containers/LivingWhitepaper/LivingWhitepaper.scss
jamessspanggg: We have extra padding on the RHS with this change

Screen Shot 2021-07-17 at 9 17 28 AM

src/containers/LivingWhitepaper/containers/PrincipalEntities/index.tsx
jamessspanggg: let me create the component from my side

src/containers/LivingWhitepaper/LivingWhitepaper.scss
thesanjeevsharma: fixing that query.

Also, there's actually more padding on RHS. I've left a comment on Figma for you.
But on smaller screens, I have kept it equal because it didn't look nice.

jamessspanggg (approved):
Amazeballs

PR #1987, opened 2021-07-24
Fix whitepaper issues, scrolling behaviour when page reload

jamessspanggg:


src/components/DataTable/DataTable.scss
```diff
@@ -3,6 +3,8 @@
 .DataTable {
   &__table {
     box-shadow: 0 1px 13px rgba(0, 0, 0, 0.06);
+    display: block;
```
Lalisfeed, line 6: @jamessspanggg it looks nice when
* `display:table;` looks much better than block
* when tables are placed inside a div as follows
```
<div  style="overflow-x:auto;"> 
<table>Table stuff</table> // remove `overflow-x: scroll;` form the table
</div>
```
what do you say?

PR #1992, opened 2021-07-25
Remove old documentation

buckyroberts:


jamessspanggg:
Are we removing everything including the wallet guide and deployment guides? I think these are not covered in the whitepaper. @manishram tagging you here as well

Screen Shot 2021-07-26 at 8 35 43 AM

buckyroberts:
@jamessspanggg we can leave those in until beta is ready since they will still be useful for users until then.

jamessspanggg:
> @jamessspanggg we can leave those in until beta is ready since they will still be useful for users until then.

I see, then should we not remove the Documentation link at the top nav? since if we remove them, then the users have no way to access unless they enter the url directly

jamessspanggg:
Maybe we can just remove the `guide` section, while leaving the others? I can takeover this PR if its ok

Screen Shot 2021-07-26 at 9 37 24 PM

manishram:
@jamessspanggg, please go ahead. :)

jamessspanggg:
@manishram updated as discussed, can check if its alright?

Screen Shot 2021-07-29 at 2 21 47 PM

manishram:
@jamessspanggg looks good to me.

PR #2009, opened 2021-08-07
Add project rules under developer portal

jamessspanggg:
Fixes part of #2001 

Note that the latest project rules is different from the existing one, with styling and wording changes

~~TODO: add project FAQ~~

PR #2011, opened 2021-08-08
Handle breadcrumb and sidemenu for projects

jamessspanggg:
Fixes part of #2009 

Pending confirmation from designer

### DIFF
--- a/src/containers/DeveloperPortalProjects/components/Breadcrumb/index.tsx
+++ b/src/containers/DeveloperPortalProjects/components/Breadcrumb/index.tsx
@@ -15,11 +15,17 @@ import './Breadcrumb.scss';
 type Props = {
   className?: string;
   breadcrumbHeight: number;
+  approvedProjectUrls?:
+    | {
+        title: string;
+        url: string;
+      }[]
+    | null;
 };
 
 const TOP_LINK_HEIGHT = 72;
 
-const Breadcrumb: FC<Props> = ({breadcrumbHeight, className}) => {
+const Breadcrumb: FC<Props> = ({approvedProjectUrls, breadcrumbHeight, className}) => {
   const location = useLocation();
   const {width} = useWindowDimensions();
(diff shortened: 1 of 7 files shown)


jamessspanggg:
@thesanjeevsharma there are still some pending confirmation from the designer about this, but i think we can merge this first and come back to it later, since ur project details page requires this.

src/containers/DeveloperPortalProjects/components/Breadcrumb/index.tsx
```diff
@@ -15,11 +15,17 @@ import './Breadcrumb.scss';
 type Props = {
   className?: string;
   breadcrumbHeight: number;
+  approvedProjectUrls?:
+    | {
+        title: string;
+        url: string;
+      }[]
+    | null;
```
thesanjeevsharma, line 23: 😮 What's this? It can be `null` or an array of objects with the shape `{title, url}`?

What does that first pipe mean?

jamessspanggg: > 😮 What's this? It can be null or an array of objects with the shape {title, url}?

Yes correct.

The first pipe is needed if your type starts in the next line, autoformat does that 😅

thesanjeevsharma:
@jamessspanggg The list doesn't close once it's opened. Can we fix this?
image

thesanjeevsharma:
Also, I cannot see the complete side menu even if I scroll. I have to scroll my page to the end to see the rest of the menu.

image

jamessspanggg:
> Also, I cannot see the complete side menu even if I scroll. I have to scroll my page to the end to see the rest of the menu.
> 
> image

yes, that's the confirmation that i require from the designer, since there will be so many projects.

jamessspanggg:
> @jamessspanggg The list doesn't close once it's opened. Can we fix this?

ok let me work on taht

jamessspanggg:
@thesanjeevsharma added the fix

thesanjeevsharma (approved):

Issue #2001, opened 2021-08-03
Integrated Projects & Rules


manishram:
**Task:**
- Need to update the Projects section (https://thenewboston.com/projects) with an updated design.

**Design:**
- https://www.figma.com/file/5madc2zFyUi7A3G0ZeFcry/Integrated-Projects-and-Rules?node-id=0%3A1

Feel free to break the task into smaller sub-tasks. Let me know if any additional information or clarification is required.

jamessspanggg:
I think we basically have to move the projects page to be under the developer portal section. @thesanjeevsharma 

The implementations should be quite similar with the existing one, except for the sidemenu where the `Approved Projects` section does not scroll accordingly. When we click one of the links in the sidemenu, it will directly visit the project details page. Only the inner subsections need to highlight and scroll accordingly.

I think we can break down the tasks to:
- [ ] Move Projects Home Page 
- [ ] Move Projects Details Page
- [ ] Move List of Projects
- [ ] Move Rules and Guidelines page

PR #2022, opened 2021-08-17
refactor: get rid of Projects container

thesanjeevsharma:
closes: #2001


src/containers/App.tsx
```diff
@@ -38,10 +37,9 @@ import TermsOfUse from './TermsOfUse';
 import Wallet from './Wallet';
 
 /**
- * Lazy load pages that may contribute a lot to the bundle size
+ * Lazy load pages that may contribute a lot to the bundle sizes
```
thesanjeevsharma, line 40: lol GitHub Co-pilot 😆

jamessspanggg:
I think we should wait for the designer's decision on the sidemenu before merging this

jamessspanggg:
@thesanjeevsharma i added a temporary fix to add the max-height to the sidemenu, do u think this solution is feasible for now? we can come back to it and fix if the designer has other opinions

thesanjeevsharma:
@jamessspanggg I think we can get rid of this scroll bar?

image

jamessspanggg (approved):

PR #2026, opened 2021-08-24
! Fix roadmap hero image position

jamessspanggg:
Screen Shot 2021-08-24 at 10 10 29 PM

### DIFF
--- a/src/containers/Roadmap/components/TeamRoadmapCard/TeamRoadmapCard.scss
+++ b/src/containers/Roadmap/components/TeamRoadmapCard/TeamRoadmapCard.scss
@@ -42,7 +42,7 @@
 
   &__content {
     max-height: 480px;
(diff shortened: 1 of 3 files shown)


src/containers/Roadmap/components/TeamRoadmapCard/TeamRoadmapCard.scss
```diff
@@ -42,7 +42,11 @@
 
   &__content {
     max-height: 480px;
-    overflow-y: scroll;
+    overflow-y: auto;
+
+    &::-webkit-scrollbar-track {
+      display: none;
+    }
```
thesanjeevsharma, line 49: @jamessspanggg I think it should work without this too. You can try on mac first, I am leaving this here.

We can remove this in some other PR, if all works fine on your end.

PR #2028, opened 2021-08-25
sfc for components

angle943:
the `container` files need them as well, but then I got lazy. We can put that in the backlog.

One thought i have had while working on the Wallet app. Styled Components uses the `className` prop to perform the magic of being able to style other StyledComponents from the reference of a parent component (https://styled-components.com/docs/basics#styling-any-component)

Let me try to explain what happened with an example. suppose we have two components:

`const Child = ({className}) => <div className={className}>...</div>`
`const Parent = () => <div><S.Child /></div>`

within the `Parent`'s `Styled` file, I add styling to the `Child` like so:

```
import UChild from './Child';

export const Child = styled(UChild)`
   background: pink;
`
```

Therefore, whenever the `<S.Child />` is rendered within the `<Parent />`, it turns pink.

So I did something like this, and after awhile, I thought (mistakenly) that `Child` component does not need the `className` component (I must've forgotten how styled-components work, but even if I did remember, it is not so obvious which component is being styled from a parent component, and which aren't), and so I removed it. The `<S.Child />`, therefore, was no longer pink. But I was unaware of this for quite awhile, until I noticed finally that the UI was busted (with pink it is obvious, but my issue was more subtle, dealing with margins).

`Styled-Components` gives no indication that there is some styling that has been defined, but because of the absense of `className`, it is not put into effect. This is quite alarming, since it may potentially mean that some styles goes unapplied without our knowledge.

One way we can solve this dilemma is to enforce that all our `components` must have an optional `className` prop, and that `className` be put within the appropriate React Element. I'm not sure if there's a linting rule that enforces this though.

But I want you guys's thoughts. Hopefully you guys understand my worry?

### DIFF
--- a/src/components/A/index.tsx
+++ b/src/components/A/index.tsx
@@ -1,13 +1,13 @@
-import React, {FC, ReactNode} from 'react';
+import React, {ReactNode} from 'react';
 import clsx from 'clsx';
 import {Icon, IconType} from '@thenewboston/ui';
 import {bemify} from '@thenewboston/utils';
+import {SFC} from 'types/generic';
 
 import './A.scss';
 
 interface ComponentProps {
   children?: ReactNode;
-  className?: string;
   dataTestId?: string;
   href: string;
   newWindow?: boolean;
@@ -15,7 +15,7 @@ interface ComponentProps {
   iconSize?: number;
 }
 
-const A: FC<ComponentProps> = ({
+const A: SFC<ComponentProps> = ({
   children,
   className,
   dataTestId = 'A',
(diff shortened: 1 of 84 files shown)


jamessspanggg:
> But I want you guys's thoughts. Hopefully you guys understand my worry?

Yep I understand the issue, I think this issue will be apparent when we migrate over to styled-components. But nonetheless, i think it's good to add this type. We can just have a mental note to add this type at every component.

jamessspanggg (approved):

thesanjeevsharma (approved):
LGTM

Yeah, I get what you're trying to solve. I think this will be helpful. 👍

Issue #2034, opened 2021-08-28
Add `SDK & Libraries` section to the Developer Portal


manishram:
**Task:**
- Add `SDK & Libraries` section to the Developer Portal

**Design:** 
- https://www.figma.com/file/6NA1usM50PE2DG74CjL8A5/SDK%2FLibraries?node-id=189%3A1969

Let me know if you need any clarification on the task.

PR #2037, opened 2021-08-30
feat/sdk and libraries

thesanjeevsharma:
closes: #2034


src/components/FormElements/Checkbox/Checkbox.scss
```diff
@@ -0,0 +1,12 @@
+.Checkbox {
+  color: var(--color-gray-700);
+  display: inline-flex;
```
jamessspanggg, line 3: btw, should we start using styled-components from now onwards?

thesanjeevsharma: after this PR, yeah 😆

src/constants/libraries.ts
```diff
@@ -0,0 +1,7 @@
+export const LANGUAGES = {
```
jamessspanggg, line 1: curious, why constant instead of enum? i believe enums are more suitable for this data type

src/containers/DeveloperPortal/containers/SdkAndLibraries/components/ConfirmationModal/ConfirmationModal.scss
```diff
@@ -0,0 +1,65 @@
+@import 'styles/font-mixins';
+
+.ConfirmationModal {
+  background-color: rgba(0, 0, 0, 0.3);
+  bottom: 0;
+  display: grid;
+  left: 0;
```
jamessspanggg, line 7: can we fix the height of the modal? as there is a slight content shift when the modal is shown.

src/containers/DeveloperPortal/containers/SdkAndLibraries/components/ConfirmationModal/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import {Button} from 'components';
+import {SFC} from 'types/generic';
+
+import LeaveSiteImg from '../../assets/leave-site.svg';
+
+import './ConfirmationModal.scss';
+
+type Props = {
+  url: string | null;
```
jamessspanggg, line 14: `url?: string` would be ok?

src/constants/libraries.ts
thesanjeevsharma: Well, in the Filter component I need all the languages. The solution I tried with enum seemed like a hack, so I made an object in constants instead. Also, extracting a `LanguageType` type from this object.

src/containers/DeveloperPortal/containers/SdkAndLibraries/components/ConfirmationModal/index.tsx
thesanjeevsharma: not sure, as the passed the `url` can have `string` or `null`; based on this I open the modal.

I thought `url?: string` works for `undefined`(more specifically, when a value is not passed in) and `string`? It works for `null` too?

src/constants/libraries.ts
jamessspanggg: > Well, in the Filter component I need all the languages.

I think you can just use `Object.Keys/Values/Entries`?
https://javascript.plainenglish.io/iterating-on-a-typescript-string-enum-45f7e36b015b

> Also, extracting a LanguageType type from this object.

If it is an `enum`, you would not need to use `keyof typeof`, you can just straight use the `enum` as the type. This is one advantage of using enum over constant

src/containers/DeveloperPortal/containers/SdkAndLibraries/components/ConfirmationModal/index.tsx
jamessspanggg: Yes `?` only accepts `undefined`

I see, in my personal opinion, if it is an optional prop, we should use `?` parameter, because to accommodate when the `url` is not passed in as props. If you set it to `string | null`, then a url needs to be always passed in.

src/types/libraries.ts
```diff
@@ -0,0 +1,12 @@
+import {LANGUAGES} from 'constants/libraries';
+
+type LanguageKeyType = keyof typeof LANGUAGES;
```
jamessspanggg, line 3: This is not needed if u use `enum`

src/components/FormElements/Checkbox/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+        totalSize={size}
+        unfocusable={unfocusable}
+      />
+      <input
+        className="Checkbox__input"
+        checked={checked}
+        disabled={disabled}
+        id={name || value}
+        name={name || value}
+        readOnly
```
jamessspanggg, line 67: should we pass the `readOnly` as props as well?

src/containers/DeveloperPortal/containers/SdkAndLibraries/components/Breadcrumb/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+            hasPrecedingArrowIcon
+            key={pathname}
+            scrollOffset={-(NAVBAR_HEIGHT + TOP_LINKS_HEIGHT + BREADCRUMB_HEIGHT)}
+            title={PATHNAME_TO_TITLE_MAPPING[pathname]}
+            titleLink={`/developer/${pathname}`}
+          />
+        );
+      })}
+      {/* Filter */}
+      {Boolean(width < 992) && (
```
jamessspanggg, line 76: Just curious, why the need to explicitly cast `width < 992` as `boolean`?

src/containers/DeveloperPortal/containers/SdkAndLibraries/components/DeveloperPortalLayout/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  pageName: string;
+  approvedProjectUrls?: {
+    title: string;
+    url: string;
+  }[];
+  projectName?: string;
+};
+
+const DeveloperPortalLayout: FC<Props> = ({children, pageName}) => {
+  const [selectedLanguages, setSelectedLanguages] = useState<string[]>([]);
```
jamessspanggg, line 23: Should this be of type `LanguageType`?

src/containers/DeveloperPortal/containers/SdkAndLibraries/components/Breadcrumb/index.tsx
thesanjeevsharma: This is out of habit. 

Earlier when I had any checks based on the length I always forgot to convert those into Boolean and then 1 would appear on screen in production. xD

So, I have made this a rule for myself, to always cast into boolean.

src/components/FormElements/Checkbox/index.tsx
thesanjeevsharma: I actually didn't write the code for this myself, it's the radio button code we had. Just changed the icon and input type.

Also, this is used for the HTML input which is hidden. Our own custom icon doesn't need to use a `readOnly` prop.

src/types/libraries.ts
thesanjeevsharma: Just before my last commit, I had it the way you are telling. Don't know why I felt this was better. :/ Let me do a force push and revert.

src/containers/DeveloperPortal/containers/SdkAndLibraries/components/Breadcrumb/index.tsx
jamessspanggg: i see hahah got it

src/containers/DeveloperPortal/containers/SdkAndLibraries/components/Breadcrumb/Breadcrumb.scss
```diff
@@ -0,0 +1,25 @@
+@import 'src/styles/mixins';
+@import 'styles/font-mixins';
+@import '../../shared';
+
+.Breadcrumb {
```
jamessspanggg, line 5: will naming it as `Breadcrumb` be conflicting to any other existing `Breadcrumb` classnames?

src/containers/DeveloperPortal/containers/SdkAndLibraries/components/DeveloperPortalLayout/DeveloperPortalLayout.scss
```diff
@@ -0,0 +1,24 @@
+@import '../../shared';
+@import 'styles/z-index';
+
+.DeveloperPortalLayout {
```
jamessspanggg, line 4: same for this, will it be conflicting with the existing `DeveloperPortalLayout`?

src/containers/DeveloperPortal/containers/SdkAndLibraries/components/TopLinks/TopLinks.scss
```diff
@@ -0,0 +1,29 @@
+@import 'styles/font-mixins';
+@import '../../shared';
+@import 'styles/z-index';
+
+.TopLinks {
```
jamessspanggg, line 5: This as well

src/types/libraries.ts
```diff
@@ -0,0 +1,15 @@
+export enum Language {
```
jamessspanggg, line 1: usually I consider an `enum` as a constant instead of a type, since it can be assigned to a variable. what do you think?

thesanjeevsharma: Yeah, but we've had enums in type files before too. 🤔 I think it's a good idea to keep these with types as our type files are already segregated and not very huge.

src/containers/DeveloperPortal/containers/SdkAndLibraries/components/Breadcrumb/Breadcrumb.scss
thesanjeevsharma: You're right! But right now, these three components have the exact same styling. So, overwriting won't break anything.

I think this will be solved automatically once we move out the dev portal or migrate to styled components.

If you've noticed any major styling issue I can rename these, otherwise we can roll with it for now.

jamessspanggg: hmm I see, if the 3 components have the exact same styling, maybe we have a common shared scss file. But yeah, maybe can be done when we migrate to styled components

jamessspanggg (approved):

PR #2063, opened 2021-09-25
Refactor urls to constants file

jamessspanggg:


src/constants/urls.ts
```diff
@@ -0,0 +1,39 @@
+// Internal URLs
```
thesanjeevsharma, line 1: `routes.ts` 😬

thesanjeevsharma (approved):

PR #2095, opened 2021-10-13
requiredAsterisk, Shadow & Slideup

arialdyscv:
closes: #2084 

converted SCSS files from these components to styled-components.


arialdyscv:
Ok @thesanjeevsharma you can check it out now. You have no idea how much i learned doing this task 😁

PR #2101, opened 2021-10-15
Convert Label, LabelFilter and HashLink to styled components

ChrisBit:
Issue #2076

Changes:
* Convert Label, LabelFilter and HashLink to styled components.
* Installs jest-styled-components to test the styled-compoents.
* Components are now styled using hex instead of rgb colors. Changed tests to reflect this. (Let me know if this is not desirable).
* Removed component names from clsx since it was only used as a selector to style the components (Correct me if i'm wrong on this). It will however still add passed classNames. I let them be because i'm not sure of what the custom classNames are actually used for.
* Changed tests to check for customClassName instead of default one. "renders with default className" -> "renders with custom className". Removed other tests that also tested only this.

### DIFF
--- a/src/components/HashLink/index.tsx
+++ b/src/components/HashLink/index.tsx
@@ -1,4 +1,3 @@
-import clsx from 'clsx';
 import React from 'react';
 import {SFC} from 'types/generic';
 import * as S from './Styles';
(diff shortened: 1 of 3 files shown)


ChrisBit:
Thanks! @jamessspanggg 😄 
I did visual, functional testing and comparements of html and css. Here are screenshots of the components:
Screenshot 2021-10-16 at 13 55 09
Screenshot 2021-10-16 at 13 41 52
Screenshot 2021-10-16 at 13 43 46

ChrisBit:
My account number: 3dd9b1f20557a76e5c42e4b8fd7fef995539b0fa16ebb0eb274fc7300fef2e43
thanks! 😄

PR #2141, opened 2021-11-09
fix: temporary solution for expiring images

jamessspanggg:


thesanjeevsharma (approved):

jamessspanggg:
Root cause is fixed by BE, no longer need this MR

Issue #2082, opened 2021-10-09
Convert Note, Pagination and Popover to styled-components


jamessspanggg:
## Task Description

Convert `Note`, `Pagination` and `Popover` component from `scss` to `styled-components`.

## To Note

Styled-component migration guidelines:
- Styled-components should be a separate file called Styles.ts
- When importing styled-components to component file, should use `import * as S from ‘./Styles’. 
- Use semantic naming for styled-components
- Try to maintain alphabetical order for css properties
- Avoid creating styled-components for HTML elements that do not require styling
- Can refer to `src/components/Avatar/Styles.ts` for examples
- Please ensure that the existing UI is not affected, then only seek for reviews

Please only work on the issues that you are assigned to, PRs will be ignored if you are not assigned to the task. If the assigned issues have not been worked on for more than 3 days, we will remove the assignment.

thesanjeevsharma:
Ah okay! My bad, didn't see it. Unassigned.

alpha-titan:
Can I work on this?

thesanjeevsharma:
> @jamessspanggg @thesanjeevsharma I would like to work on this issue. Can anyone of you assign me to this issue?

@lokeshn011101 since you've already made a contribution for hacktoberfest, we cannot assign another hacktoberfest issue to you until 25th Oct. But feel free to pick up other issues; you can still earn bounty. :)

gauravwadhwani:
Can I work on this?

Creatinas:
Can i take this one? (It would be my second first-timers-only)

thesanjeevsharma:
@Creatinas go ahead, no problem.

wakawakathedev:
@Creatinas is this PR ready to be reviewed?

manishram:
@Creatinas what is your TNB Account Number?

Creatinas:
@manishram 679aff9a630beb883cc10368a2dedb7722e63b682143ae9fd2e051c13b710752

PR #2164, opened 2021-12-11
[#2082] Refactored Pagination, Popover components to styled-components

Creatinas:
closes: #2082 

Refactored Pagination, Popover components to styled-components and removed Notes component, not used.


wakawakathedev:
Ref #2082

PR #2187, opened 2022-02-09
Utilise CSS sprites for logos to prevent overhead to fetch multiple images

jamessspanggg:
## Background

Currently in pages such as https://thenewboston.com/social, the page issues a HTTP request for each of the logos, which 
might cause unnecessary overhead (there are more than 10 logos). We can instead use CSS sprites to reduce the number of HTTP requests to a single one.

## Changes

Added dependency for [spritesmith-cli](https://www.npmjs.com/package/spritesmith-cli) to achieve this use case. Also reduced the size of some logos as they were unnecessarily big in size.

Running `npm run css-sprite` will generate the `logos-sprite.png` image, as well as the `MarketingCard/sprite.scss` file

## Before

Screen Shot 2022-02-09 at 9 54 40 PM

## After

Screen Shot 2022-02-09 at 9 54 51 PM

References: https://css-tricks.com/css-sprites/


src/components/MarketingCard/index.tsx
```diff
@@ -4,45 +4,21 @@ import {SFC} from 'types/generic';
...
 
 export interface MarketingCardProps {
   customLink?: string;
   website: SocialMedia;
 }
 
 const MarketingCard: SFC<MarketingCardProps> = ({customLink, website}) => {
-  const renderImage = (src: any) => <S.Image alt={website} data-testid="MarketingCard__img" src={src} />;
-
-  const images = {
```
jamessspanggg, line 29: removed this ineffecient implementation, as every call to `MarketingCard` will render all logo images, whereas we can implement it such that each marketing card renders a single image

PR #2204, opened 2022-07-19
tnb

buckyroberts:


src/containers/App/index.tsx
```diff
@@ -0,0 +1,15 @@
...
+
+import * as S from './Styles';
+
+const App: FC = () => {
+  return (
+    <S.Container>
+      <S.Unite>
+        U<S.Red>n</S.Red>ite.
+      </S.Unite>
+    </S.Container>
```
thesanjeevsharma, line 11: 👀 Okay!

nzkozar:
What is the goal of this? What happened?

